PR #14, opened 2017-07-09
alicloud/provider: Add a new alicloud resource oss_bucket_object

xiaozhu36:
This PR add a new alicloud resource oss bucket object, and it contains oss bucket object's resource provider, test cases and introduction documents.

The PR has a dependency on terraform-providers/terraform-provider-alicloud#10.

The running results of oss bucket object test cases as follows:

TF_ACC=1 go test ./alicloud/ -v -run=TestAccAlicloudOssBucketObject -timeout=120m
=== RUN   TestAccAlicloudOssBucketObject_source
--- PASS: TestAccAlicloudOssBucketObject_source (16.41s)
=== RUN   TestAccAlicloudOssBucketObject_content
--- PASS: TestAccAlicloudOssBucketObject_content (16.13s)
=== RUN   TestAccAlicloudOssBucketObject_acl
--- PASS: TestAccAlicloudOssBucketObject_acl (14.30s)
PASS
ok      github.com/terraform-providers/terraform-provider-alicloud/alicloud     46.864s


alicloud/resource_alicloud_oss_bucket_object.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	options, err := buildObjectHeaderOptions(d)
+	if err != nil {
+		return fmt.Errorf("Error building object header options: %#v", err)
+	}
+
+	object, err := bucket.GetObjectDetailedMeta(d.Get("key").(string), options...)
+	if err != nil {
+		return fmt.Errorf("Error Reading Object: %#v", err)
```
catsby: Is it possible this call will return an `err` that represents a `404 not found`, such that we would return here and not get to check if the `object` is `nil` in the next block, where we would also call `d.SetId("")`?

xiaozhu36:
@catsby Thanks for your review.

I have fixed the conflict and please review again, thanks a lot

Regards,
Guimin He

xiaozhu36:
@catsby Thanks for your review.

I have fixed the conflict and please review again, thanks a lot

Regards,
Guimin He

PR #18, opened 2017-08-09
alicloud/provider:	add new resource container cluster

xiaozhu36:
The PRs mainly add a new resource alicloud_container_cluster. The PR includes the resource code, test cases, examples and docs.

The running results of its test case as follows:

TF_ACC=1 go test ./alicloud -v -run=TestAccAlicloudContainerCluster -timeout=120m
=== RUN   TestAccAlicloudContainerCluster_importBasic
--- PASS: TestAccAlicloudContainerCluster_importBasic (439.56s)
=== RUN   TestAccAlicloudContainerCluster_classic
--- PASS: TestAccAlicloudContainerCluster_classic (435.17s)
=== RUN   TestAccAlicloudContainerCluster_vpc
--- PASS: TestAccAlicloudContainerCluster_vpc (716.03s)

PASS
ok github.com/alibaba/terraform-provider/alicloud  1590.794s


alicloud/resource_alicloud_container_cluster.go
```diff
@@ -0,0 +1,253 @@
...
+
+	cluster, err := conn.DescribeCluster(d.Id())
+
+	if err != nil {
+		return err
+	}
+
+	d.Set("name", cluster.Name)
+	d.Set("size", cluster.Size)
+	d.Set("network_mode", cluster.NetworkMode)
```
catsby: I don't see this attribute defined in the schema

alicloud/resource_alicloud_container_cluster.go
```diff
@@ -0,0 +1,253 @@
...
+	if cluster.VPCID != "" {
+		d.Set("vpc_id", cluster.VPCID)
+	}
+	if cluster.VSwitchID != "" {
+		d.Set("vswitch_id", cluster.VSwitchID)
+	}
+
```
catsby: It seems there are some fields that are not re-read from the API here, specifically:
- `cidr_block`
- `instance_type`
- `disk_size`
- `disk_category`
- `image_id`

Not reading these attributes from the API will prevent Terraform from detecting drift in the configuration, or otherwise ensuring that Terraform did create the resource as desired. Does the API return this information?

xiaozhu36: Hi @catsby 

Thanks for your review. The API doesn't return above fields and if them have been changed, the resource would be re-create. I can get their value from schema and return. Could please provide better advice? Thanks.

Regards,
Guimin

PR #65, opened 2018-01-04
modify db_instance and output more useful error message

xiaozhu36:
The pr modify db_instance resource and remove some field to next new resources. In addition, the PR output much useful message when there are some errors.

The result of running test case as follows:

TF_ACC=1 go test ./alicloud -v -run=TestAccAlicloudDBInstance -timeout 120m
=== RUN   TestAccAlicloudDBInstance_import
--- PASS: TestAccAlicloudDBInstance_import (548.36s)
=== RUN   TestAccAlicloudDBInstance_basic
--- PASS: TestAccAlicloudDBInstance_basic (451.32s)
=== RUN   TestAccAlicloudDBInstance_vpc
--- PASS: TestAccAlicloudDBInstance_vpc (470.51s)
=== RUN   TestAccAlicloudDBInstance_multiAZ
--- PASS: TestAccAlicloudDBInstance_multiAZ (473.04s)
=== RUN   TestAccAlicloudDBInstance_securityIps
--- PASS: TestAccAlicloudDBInstance_securityIps (489.98s)
=== RUN   TestAccAlicloudDBInstance_upgradeClass
--- FAIL: TestAccAlicloudDBInstance_upgradeClass (1799.57s)
PASS
ok    github.com/alibaba/terraform-provider/alicloud  4233.446s


catsby:
Hello! I have a few concerns about this pull request that I'd like to discuss.

1) The title is misleading in that while those changes are made, they represent a very small fraction of the actual changes contained. In the future please construct pull requests to isolate specific changes
2) Those db instance changes appear to be significant and perhaps even backwards incompatible changes. For example, if a user is on a version of the provider that does not contain these changes, and then upgrades, will their next Terraform `plan` show a diff? Will it show any `ForceNew` (recreation of resources) needed? These things need to be considered and addressed in upgrades and state migrations
3) The changes to `config.go` and `provider.go` are backwards incompatible. The introduce a new configuration attribute `SecurityToken` (`security_token`) that is now **required**, meaning any user upgrading to this version will now have configuration that doesn't work. If this new attribute is a new requirement and so unavoidable, it needs to be documented in both the CHANGELOG as a backwards incompatible change, as well as documentation added for what this new value is and that it is required (in /website/docs/index.html.markdown)
4) Lastly, I see there are new versions/releases in GitHub (https://github.com/terraform-providers/terraform-provider-alicloud/releases). These new releases are not automatically available to Terraform users. For those versions to become available to all Terraform users, they need to be cross compiled and hosted on releases.hashicorp.com. The process to publish releases is controlled by HashiCorp at this time, and new releases need to be coordinated through us. Do you have an email address contact that you can message to start this process? 

Please let me know what questions you have here. It's clear we have not provided you the information you need, so I would very much like to clear things up and get your new releases out 😄 

Thanks!

zhuzhih2017:
@catsby Thanks for you comments, I need your help to make our new release out. My email dongxiao.zzh@alibaba-inc.com

1. I'll pay attentions for  constructing pull requests to isolate specific changes in the future please.
2. We have do the compatible testing and we keep the old field to make sure the not broken the state and add annotation to prompt user when they change. 
3. You are right, SecurityToken shouldn't be required. We have fixed.

PR #98, opened 2018-01-25
Fix implicit regex/string conversion issue

catsby:
There is currently an issue that prevents the AliCloud documentation from rendering. Please merge this as soon as possible (or I may) to restore the documentation.

### DIFF
--- a/website/alicloud.erb
+++ b/website/alicloud.erb
@@ -10,7 +10,7 @@
                     <a href="/docs/providers/alicloud/index.html">Alicloud Provider</a>
                 </li>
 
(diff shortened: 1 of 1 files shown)


catsby:
The AliCloud documentation is currently missing from terraform.io due to this rendering error, so in the interest of restoring them, I'm going to go ahead and merge and deploy this patch. Let me know if you have any questions. Thanks!

xiaozhu36:
Hi @catsby Thanks for your contribution. I will avoid the same issue to happen in the future PR.

PR #126, opened 2018-03-14
Update SDK version to fix unresolving endpoint issue

xiaozhu36:
The PR updates sdk version to fix bug that provider cannot resolving some region endpoint.


zhuzhih2017:
@xiaozhu36 Please keep the same revision for the SDK from same repository.

PR #135, opened 2018-03-26
new data source for security group rules

malferov:
- new data source added to provider for security group rules
 - test case for filtering different types of rules
 - documentation page


alicloud/data_source_alicloud_security_group_rules.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/denverdino/aliyungo/ecs"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAlicloudSecurityGroupRules() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAlicloudSecurityGroupRulesRead,
+
+		Schema: map[string]*schema.Schema{
+			"id": {
```
xiaozhu36: I think 'group_id' may be more easily understandable than 'id'. 'name' and 'description' are also same.  After all, the datasource is security group rules, not security group.

malferov: Indeed, fixed. Please find update.

PR #368, opened 2018-09-29
add node_cidr_mas and log_config, fix worker_data_disk issue

xh4n3:


alicloud/resource_alicloud_cs_kubernetes.go
```diff
@@ -475,9 +505,31 @@ func resourceAlicloudCSKubernetesRead(d *schema.ResourceData, meta interface{})
...
+		} else {
+			logConfig["type"] = cluster.Parameters.LoggingType
+			if cluster.Parameters.SLSProjectName == "None" {
+				logConfig["sls_project_name"] = ""
+			} else {
+				logConfig["sls_project_name"] = cluster.Parameters.SLSProjectName
+			}
+			if err := d.Set("log_config", []map[string]interface{}{logConfig}); err != nil {
+				return err
+			}
```
xiaozhu36: I think if LoggingType == "None" and SLSProjectName == "None", you can set log_config to a empty struct directly; otherwise set logtype and project.

alicloud/resource_alicloud_cs_kubernetes.go
```diff
@@ -778,21 +832,35 @@ func buildKubernetesArgs(d *schema.ResourceData, meta interface{}) (*cs.Kubernet
...
+	}
+
+	if v, ok := d.GetOk("worker_data_disk_category"); ok {
+		creationArgs.WorkerDataDiskCategory = v.(string)
+		creationArgs.WorkerDataDisk = true
+		if size, ok := d.GetOk("worker_data_disk_size"); ok {
+			creationArgs.WorkerDataDiskSize = int64(size.(int))
+		} else {
+			creationArgs.WorkerDataDiskSize = 40
+		}
```
xiaozhu36: I think both `worker_data_disk_category` and `worker_data_disk_size` must be set at the same time, or do not.

alicloud/resource_alicloud_cs_kubernetes.go
```diff
@@ -149,6 +155,32 @@ func resourceAlicloudCSKubernetes() *schema.Resource {
...
+			"log_config": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:         schema.TypeString,
+							ValidateFunc: validateAllowedStringValue([]string{KubernetesClusterLoggingTypeSLS}),
```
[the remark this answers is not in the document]
xh4n3: I set `type` to be optional because `type` can be either "" or "SLS"

alicloud/resource_alicloud_cs_kubernetes.go
xh4n3: Conditions here do have issues.
I will set `type` when `type` is not None, no matter if `project` exists.

alicloud/resource_alicloud_cs_kubernetes.go
xh4n3: worker_data_disk_size are 40 by default when worker_data_disk_category is set.

PR #389, opened 2018-10-09
add new resource of apigateway apis and groups

geeklenny:
add new resource of apigateway apis and groups


geeklenny:
add new resource of apigateway apis and groups

PR #471, opened 2018-11-05
support enable/disable security enhancement strategy when create inst…

nightmeng:
TF_DEBUG=debug TF_ACC=1 go  test -v -run=TestAccAlicloudInstance_basic
=== RUN   TestAccAlicloudInstance_basic
--- PASS: TestAccAlicloudInstance_basic (104.59s)
PASS
ok  	github.com/terraform-providers/terraform-provider-alicloud/alicloud	104.641s

### DIFF
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -485,3 +485,7 @@ func loadFileContent(v string) ([]byte, error) {
 	}
 	return fileContent, nil
 }
+
+func equalsIgnoreCase(left string, right string) bool {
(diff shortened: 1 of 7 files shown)


alicloud/common.go
```diff
@@ -485,3 +485,7 @@ func loadFileContent(v string) ([]byte, error) {
 	}
 	return fileContent, nil
 }
+
+func equalsIgnoreCase(left string, right string) bool {
```
xiaozhu36: I think use the ValidateFunc will be better.

PR #481, opened 2018-11-08
Increasing the min_size and max_size scaling group parameters to 1000

agrotarun:
Currently the max capacity of auto scaling in ALI cloud is 1000 but terraform still points to old capacity of 100. Changing it


xiaozhu36:
HI @agrotarun Thanks for your contribution. Please update docs for this PR.

PR #521, opened 2018-11-23
fixed bug in backup policy update

arafato:
If backup policy is updated there was an error since required parameters where not set.
Also fixed docs: backup policy input parameters have been aligned with resource provider names.

### DIFF
--- a/alicloud/resource_alicloud_kvstore_backup_policy.go
+++ b/alicloud/resource_alicloud_kvstore_backup_policy.go
@@ -94,24 +94,14 @@ func resourceAlicloudKVStoreBackupPolicyRead(d *schema.ResourceData, meta interf
 }
 
 func resourceAlicloudKVStoreBackupPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*connectivity.AliyunClient)
-	update := false
-	request := r_kvstore.CreateModifyBackupPolicyRequest()
-	request.InstanceId = d.Id()
-
-	if d.HasChange("backup_time") {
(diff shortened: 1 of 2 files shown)


alicloud/resource_alicloud_kvstore_backup_policy.go
```diff
@@ -98,16 +98,12 @@ func resourceAlicloudKVStoreBackupPolicyUpdate(d *schema.ResourceData, meta inte
...
-		update = true
-	}
-
-	if d.HasChange("backup_period") {
-		periodList := expandStringList(d.Get("backup_period").(*schema.Set).List())
-		backupPeriod := fmt.Sprintf("%s", strings.Join(periodList[:], COMMA_SEPARATED))
-		request.PreferredBackupPeriod = backupPeriod
+	if d.HasChange("backup_time") || d.HasChange("backup_period") {
 		update = true
 	}
```
xiaozhu36: I think `update` can be removed and the update method can be modified as following:
```
func resourceAlicloudKVStoreBackupPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
	if d.HasChange("backup_time") || d.HasChange("backup_period") {
		request := r_kvstore.CreateModifyBackupPolicyRequest()
	        request.InstanceId = d.Id()
	        request.PreferredBackupTime = d.Get("backup_time").(string)
                ....
                if _, err := client.WithRkvClient(func(rkvClient *r_kvstore.Client) (interface{}, error) {
 			return rkvClient.ModifyBackupPolicy(request)
 		}); err != nil {
 			return err
 		}
	}

PR #613, opened 2019-01-03
Use HTTPS for all requests

chanind:
Attempting to use HTTPS everywhere to address https://github.com/terraform-providers/terraform-provider-alicloud/issues/609


chanind:
It seems like you've fixed this, so closing

PR #614, opened 2019-01-03
Add capability for slb http listener redirection

Savasw:


alicloud/resource_alicloud_slb_listener.go
```diff
@@ -100,6 +100,19 @@ func resourceAliyunSlbListener() *schema.Resource {
...
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+			},
+			//http
+			"forward_port": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
```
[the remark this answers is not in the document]
xiaozhu36: And, forward_port's also should be ignore when listener_forward is false

alicloud/resource_alicloud_slb_listener.go
```diff
@@ -369,6 +397,10 @@ func resourceAliyunSlbListenerCreate(d *schema.ResourceData, meta interface{}) e
 		return fmt.Errorf("WaitForListener %s got error: %#v", Running, err)
 	}
 
+	if isHttpRedirected {
+		return resourceAliyunSlbListenerRead(d, meta)
+	}
+
```
xiaozhu36: Why need isHttpRedirected?

alicloud/resource_alicloud_slb_listener.go
```diff
@@ -399,6 +431,14 @@ func resourceAliyunSlbListenerUpdate(d *schema.ResourceData, meta interface{}) e
...
 	protocol := Protocol(d.Get("protocol").(string))
 
+	//if redirection is enabled, set http listener attribute is not supported
+	if protocol == Http {
+		listenerForward := d.Get("listener_forward")
+		if listenerForward.(bool) {
+			return fmt.Errorf("SetLoadBalancer%sListenerAttribute is not allowed when redirection is enabled.", strings.ToUpper(string(protocol)))
+		}
+	}
+
```
xiaozhu36: I think line 434-441 are needless.

alicloud/resource_alicloud_slb_listener.go
```diff
@@ -399,6 +431,14 @@ func resourceAliyunSlbListenerUpdate(d *schema.ResourceData, meta interface{}) e
...
 	d.Partial(true)
 
 	commonArgs := buildListenerCommonArgs(d, meta)
```
xiaozhu36: Can you add a testcase for this pr?

alicloud/resource_alicloud_slb_listener.go
Savasw: with http redirected to https, the SetLoadBalancerHttpListenerAttribute is not supported, hence i have used that bool var to directly go for read. Please suggest if i have missed something in the approach.

alicloud/resource_alicloud_slb_listener.go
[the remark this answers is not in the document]
Savasw: I have addressed this

alicloud/resource_alicloud_slb_listener.go
Savasw: @xiaozhu36 i have looked into adding the test case for this PR and figured that the config keys like bandwidth, backend_port are not required for redirection scenario but with current implementation for other types/scenarios they are required and also x-forwarded-for key is set to computed but here it is not valid. So what would be the suggested approach to this?

alicloud/resource_alicloud_slb_listener.go
xiaozhu36: I think there is no need to set isHttpRedirected because of listener_forward and forward_port only can be effect when creating a new http listener. After that, there is no any change for them in the update method.

Savasw: @xiaozhu36 actually the way this resource is implemented is that after create, the update is also called and there are few defaults which then make update true and it goes for SetLoadBalancerHttpListenerAttribute api. Initially i had not used this bool but after getting the error, i used this bool to bypass update when creating for first time.

xiaozhu36: So, if one listener is set HttpRedirected, it can not be updated whatever whether it is a new resource. Right?

Savasw: > So, if one listener is set HttpRedirected, it can not be updated whatever whether it is a new resource. Right?

Yes, that's correct. Update is not supported for http redirected listener

xiaozhu36: ok, can you add a note for this limitation?

alicloud/resource_alicloud_slb_listener.go
xiaozhu36: If removing `compute`, do all of listener testcases pass?

alicloud/resource_alicloud_slb_listener.go
Savasw: @xiaozhu36 where should i put that note? i had kept a condition in update to give descriptive error to user but removed it as they were needless in a way.

alicloud/resource_alicloud_slb_listener.go
Savasw: all testcase pass because all have x_forwarded_for configured. If we will remove computed, then it might create issue when x_forwarded_for is not configured as it is not mandatory. Will have to check that scenario in real use case

Savasw: @xiaozhu36 following is one workaround that would be required to be done in config to avoid the x_forwarded_for key issue, user will have to add x_forwarded_for key in config even when redirected. i have tested this.
```
resource "alicloud_slb_listener" "http" {
  load_balancer_id = "${alicloud_slb.instance.id}"
  backend_port = 80
  frontend_port = 80
  protocol = "http"
  health_check = "off"
  listener_forward = true
  forward_port = 443
  bandwidth = 10
  depends_on = ["alicloud_slb_listener.https"]
  x_forwarded_for = {
    retrive_slb_ip = false
    retrive_slb_id = false
	retrive_slb_proto = false
  }
}
```

here is the output of all
```
go test ./alicloud -v -run=TestAccAlicloudSlbListener -timeout=1440m         === RUN   TestAccAlicloudSlbListenersDataSource_http
--- PASS: TestAccAlicloudSlbListenersDataSource_http (50.56s)
=== RUN   TestAccAlicloudSlbListenersDataSource_https
--- PASS: TestAccAlicloudSlbListenersDataSource_https (27.11s)
=== RUN   TestAccAlicloudSlbListenersDataSource_tcp
--- PASS: TestAccAlicloudSlbListenersDataSource_tcp (45.61s)
=== RUN   TestAccAlicloudSlbListenersDataSource_udp
--- PASS: TestAccAlicloudSlbListenersDataSource_udp (48.03s)
=== RUN   TestAccAlicloudSlbListenersDataSource_empty
--- PASS: TestAccAlicloudSlbListenersDataSource_empty (39.19s)
=== RUN   TestAccAlicloudSlbListener_importHttp
--- PASS: TestAccAlicloudSlbListener_importHttp (19.47s)
=== RUN   TestAccAlicloudSlbListener_importHttps
--- PASS: TestAccAlicloudSlbListener_importHttps (21.06s)
=== RUN   TestAccAlicloudSlbListener_importTcp
--- PASS: TestAccAlicloudSlbListener_importTcp (17.52s)
=== RUN   TestAccAlicloudSlbListener_importUdp
--- PASS: TestAccAlicloudSlbListener_importUdp (18.06s)
=== RUN   TestAccAlicloudSlbListener_http
--- PASS: TestAccAlicloudSlbListener_http (20.21s)
=== RUN   TestAccAlicloudSlbListener_http_redirection
--- PASS: TestAccAlicloudSlbListener_http_redirection (29.04s)
=== RUN   TestAccAlicloudSlbListener_https
--- PASS: TestAccAlicloudSlbListener_https (30.83s)
=== RUN   TestAccAlicloudSlbListener_https_shared_performance
--- PASS: TestAccAlicloudSlbListener_https_shared_performance (22.03s)
=== RUN   TestAccAlicloudSlbListener_tcp
--- PASS: TestAccAlicloudSlbListener_tcp (16.95s)
=== RUN   TestAccAlicloudSlbListener_tcp_server_group
--- PASS: TestAccAlicloudSlbListener_tcp_server_group (194.20s)
=== RUN   TestAccAlicloudSlbListener_udp
--- PASS: TestAccAlicloudSlbListener_udp (17.39s)
PASS
ok      github.com/terraform-providers/terraform-provider-alicloud/alicloud     617.277s

```
if this is good for you, then i can commit the test case.

xiaozhu36:
Duplicated with #981

PR #783, opened 2019-02-21
fix default route table id

ala2208:
Hi,
The PR fixes the issue "alicloud_vpc.vpc_name.route_table_id returns custom route table id, but not the default one"
Link https://github.com/terraform-providers/terraform-provider-alicloud/issues/777

Pls. validate

### DIFF
--- a/alicloud/resource_alicloud_vpc.go
+++ b/alicloud/resource_alicloud_vpc.go
@@ -122,29 +122,34 @@ func resourceAliyunVpcRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("description", resp.Description)
 	d.Set("router_id", resp.VRouterId)
 	request := vpc.CreateDescribeVRoutersRequest()
+	requestRouteTables := vpc.CreateDescribeRouteTablesRequest()
 	request.RegionId = client.RegionId
 	request.VRouterId = resp.VRouterId
-	var response vpc.DescribeVRoutersResponse
+
+	var responseRouteTables vpc.DescribeRouteTablesResponse
 	if err := resource.Retry(6*time.Minute, func() *resource.RetryError {
 		raw, e := client.WithVpcClient(func(vpcClient *vpc.Client) (interface{}, error) {
-			return vpcClient.DescribeVRouters(request)
+			return vpcClient.DescribeRouteTables(requestRouteTables)
 		})
+
 		if e != nil && IsExceptedErrors(e, []string{Throttling}) {
 			time.Sleep(10 * time.Second)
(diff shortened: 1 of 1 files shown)


alicloud/resource_alicloud_vpc.go
```diff
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/denverdino/aliyungo/ecs"
```
xiaozhu36: I don't suggest to use the sdk "github.com/denverdino/aliyungo/ecs" and it will be removed from provider.

alicloud/resource_alicloud_vpc.go
```diff
@@ -122,27 +123,38 @@ func resourceAliyunVpcRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("description", resp.Description)
 	d.Set("router_id", resp.VRouterId)
 	request := vpc.CreateDescribeVRoutersRequest()
+	requestRouteTables := vpc.CreateDescribeRouteTablesRequest()
```
xiaozhu36: I think request can be instead requestRouteTables

alicloud/resource_alicloud_vpc.go
```diff
@@ -122,27 +123,38 @@ func resourceAliyunVpcRead(d *schema.ResourceData, meta interface{}) error {
...
-		d.Set("route_table_id", response.VRouters.VRouter[0].RouteTableIds.RouteTableId[0])
-	} else {
+
+	foundDefaultVpcRouteTableId := false
+	for _,tbl :=  range responseRouteTables.RouteTables.RouteTable {
+
+		if tbl.VRouterId == resp.VRouterId && ecs.RouteTableType(tbl.RouteTableType) == ecs.RouteTableSystem {
+			d.Set("router_table_id", tbl.RouteTableId)
+			d.Set("route_table_id", tbl.RouteTableId)
+			foundDefaultVpcRouteTableId = true
```
xiaozhu36: I think foundDefaultVpcRouteTableId is redundant.

alicloud/resource_alicloud_vpc.go
ala2208: What do you suggest to use instead of the sdk "github.com/denverdino/aliyungo/ecs"?

alicloud/resource_alicloud_vpc.go
ala2208: not sure what you mean by that, can you share an example?

xiaozhu36:
same as #800

PR #1010, opened 2019-04-16
`resource group` support for creating new VPC

pingz:
Add new `resource_group_id` parameter, creating VPC for the specific resource group.

example:

```provider "alicloud" {}

resource "alicloud_vpc" "vpc" {
  name       = "example-vpc"
  cidr_block = "192.168.0.0/16"
  resource_group_id = "rg-[resource-group-id]"
}```

### DIFF
--- a/alicloud/resource_alicloud_vpc.go
+++ b/alicloud/resource_alicloud_vpc.go
@@ -29,6 +29,10 @@ func resourceAliyunVpc() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validateCIDRNetworkAddress,
 			},
+			"resource_group_id": {
+				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


xiaozhu36:
HI @pingz Thanks for you contribution. Can you add a testcase for this PR?

pingz:
@xiaozhu36 thanks for your comment, for an acceptance testing the catch is to my knowledge for now the API for `resource group` is not public available, so I can't write a test case without manually setup a resource group which is not a viable idea for a public codebase.

I'm merely a Aliyun user, so my main interesting is merely using resource group to limit the risk of testing my terraform script as mentioned in the issue #1011.

xiaozhu36:
Got  it. Thanks a lot.

PR #1047, opened 2019-04-22
Fix pvtz_zone_attachment vpcId is not exists bug

YY-tomorrow:


kamalika-majumder:
Hi ,
Is this fix available yet. I would like to use this as I am facing the exact same issue and its causing my setup to fail intermittently.

PR #1148, opened 2019-05-14
improve slb_attachment resource use common method

YY-tomorrow:


alicloud/resource_alicloud_slb_test.go
```diff
@@ -95,7 +95,7 @@ func TestAccAlicloudSlb_classictest(t *testing.T) {
 	ra := resourceAttrInit(resourceId, nil)
 	rc := resourceCheckInitWithDescribeMethod(resourceId, &v, func() interface{} {
 		return &SlbService{testAccProvider.Meta().(*connectivity.AliyunClient)}
-	}, "DescribeSLB")
+	}, "DescribeSlb")
```
xiaozhu36: I think there is no need to specify the method

PR #1204, opened 2019-05-24
support set oss bucket stroage class

huiguangjun:


alicloud/resource_alicloud_oss_bucket.go
```diff
@@ -204,8 +204,10 @@ func resourceAlicloudOssBucket() *schema.Resource {
 				Computed: true,
 			},
 			"storage_class": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:         schema.TypeString,
+				Default:      oss.StorageStandard,
+				Optional:     true,
```
xiaozhu36: Can the storage_class be modified? If not, please set `ForceNew` to the parameter.

alicloud/validators.go
```diff
@@ -754,6 +754,18 @@ func validateOssBucketObjectServerSideEncryption(v interface{}, k string) (ws []
...
+			errors = append(errors, fmt.Errorf(
+				"%q must be a valid  StorageClass value , expected %s, %s or %s, got %q",
+				k, oss.StorageStandard, oss.StorageIA, oss.StorageArchive, strorage))
+		}
+	}
+	return
+}
+
 func validateDomainName(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
```
xiaozhu36: Please move the parameter `storage_class ` to `Argument Reference` in the oss bucket docs.

alicloud/resource_alicloud_oss_bucket.go
huiguangjun: add the ForceNew to the storage_class parameter

alicloud/validators.go
huiguangjun: done.

Issue #1072, opened 2019-04-25
Feature request: setting RAM account password policy


programmer04:
There is no resource for setting password requirements in Alibaba account. Could you add something similar like a resource [aws_iam_account_password_policy](https://www.terraform.io/docs/providers/aws/r/iam_account_password_policy.html) which exists in AWS provider for Alicloud?

programmer04:
Resolved with merging of my PR

PR #1212, opened 2019-05-27
Add new resource alicloud_ram_account_password_policy

programmer04:
This PR resolves issue #1072 
It adds a new resource `alicloud_ram_account_password_policy` with tests and documentation.


alicloud/resource_alicloud_ram_account_password_policy.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if v, ok := d.GetOk("require_symbols"); ok {
+		request.RequireSymbols = requests.NewBoolean(v.(bool))
+	} else {
+		request.RequireSymbols = requests.NewBoolean(false)
+	}
+	if v, ok := d.GetOk("max_login_attempts"); ok {
+		request.MaxLoginAttemps = requests.NewInteger(v.(int))
+	} else {
+		request.MaxLoginAttemps = requests.NewInteger(0)
```
Lexsss: I think it should be 5.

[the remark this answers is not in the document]
programmer04: Value 5 is the default for `MaxLoginAttemps` (for 5 condition will be evaluated as true). In this piece of code, I provide a possibility for assigning 0 to this property (valid range is [0-32], where 0 means disable). I have to do it in this way due to the limitation of `GetOk()` (treats false, 0, etc. value the same as empty) with GetOkExists() it would be more elegant, but one's not available.

alicloud/resource_alicloud_ram_account_password_policy.go
```diff
@@ -0,0 +1,222 @@
...
+
+func resourceAlicloudRamAccountPasswordPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*connectivity.AliyunClient)
+
+	request := ram.CreateGetPasswordPolicyRequest()
+	raw, err := client.WithRamClient(func(ramClient *ram.Client) (interface{}, error) {
+		return ramClient.GetPasswordPolicy(request)
+	})
+	if err != nil {
+		return WrapErrorf(err, DefaultErrorMsg, "alicloud_ram_account_password_policy", request.GetActionName(), AlibabaCloudSdkGoERROR)
```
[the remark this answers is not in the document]
programmer04: I think that there is no such possibility that this policy isn't available. SDK (API) supports:
- [SetPasswordPolicy](https://api.aliyun.com/?spm=a2c4g.11186623.2.12.7420368aw3tzdr#/?product=Ram&api=SetPasswordPolicy&params={}&tab=DEMO&lang=GO)
- [GetPasswordPolicy](https://api.aliyun.com/?spm=a2c4g.11186623.2.12.3e9b368aX0AzNe#/?product=Ram&api=GetPasswordPolicy&params={}&tab=DEMO&lang=GO)

The account always has this policy sets with default values. User can't delete it, only change it.

website/docs/r/ram_account_password_policy.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+* `default_require_numbers` - (Optional, Type: bool) Specifies if the occurrence of a number in the password is mandatory. Default to true.
+* `default_require_symbols` - (Optional, Type: bool) Specifies if the occurrence of a special character in the password is mandatory. Default to true.
+* `default_hard_expiry` - (Optional, Type: bool) Specifies if a password can expire in a hard way. Default to false.
+* `default_max_password_age` - (Optional, Type: int) The number of days after which password expires. A value of 0 indicates that the password never expires. Valid value range: [0-1095]. Default to 0.
+* `default_password_reuse_prevention` - (Optional, Type: int) User is not allowed to use the latest number of passwords specified in this parameter. A value of 0 indicates the password history check policy is disabled. Valid value range: [0-24]. Default to 0.
+* `default_max_login_attempts` - (Optional, Type: int) Maximum logon attempts with an incorrect password within an hour. Valid value range: [0-32]. Default to 5.
```
Lexsss: Maybe Attributes Reference and Import here.

programmer04: Thanks for the advice. I added info about `Import`. I don't need `Attributes reference`, because there aren't new exported values in addition to the arguments listed in `Argument Reference`.

alicloud/resource_alicloud_ram_account_password_policy.go
```diff
@@ -0,0 +1,222 @@
...
+
+	request := ram.CreateSetPasswordPolicyRequest()
+	if v, ok := d.GetOk("minimum_password_length"); ok {
+		request.MinimumPasswordLength = requests.NewInteger(v.(int))
+	}
+	// due to lack of GetOkExists() in old Terraform included in vendor dir
+	// GetOk() doesn't distinguish beetwen false/0 and empty, so in else clause set 0 or false
+	if v, ok := d.GetOk("require_lowercase_characters"); ok {
+		request.RequireLowercaseCharacters = requests.NewBoolean(v.(bool))
+	} else {
```
Lexsss: What do you think if we don't have `else` in all update? Maybe it's not necessary.

programmer04: Thanks for the advice, everything is always set, so I can change from `GetOk()` to `Get()` and streamline logic

PR #1239, opened 2019-05-31
fix cr repo details, improve cs and cr docs

xh4n3:


xh4n3:
the cr_repos has an issue where details are exported if we left enable_details unset.

PR #1246, opened 2019-06-03
Set default retrive_slb_proto to true

xuqingfeng:
In [slb_listener docs](https://github.com/terraform-providers/terraform-provider-alicloud/blob/master/website/docs/r/slb_listener.html.markdown#block-x_forwarded_for) it's said `retrive_slb_proto` will be set to `true` by default.

But in the actual [code](https://github.com/terraform-providers/terraform-provider-alicloud/blob/master/alicloud/resource_alicloud_slb_listener.go#L259), it's set to `false`.

This PR will fix this issue.

### DIFF
--- a/alicloud/resource_alicloud_slb_listener.go
+++ b/alicloud/resource_alicloud_slb_listener.go
@@ -256,7 +256,7 @@ func resourceAliyunSlbListener() *schema.Resource {
 						"retrive_slb_proto": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							Default:  false,
+							Default:  true,
 						},
 					},
 				},


alicloud/resource_alicloud_slb_listener.go
```diff
@@ -256,7 +256,7 @@ func resourceAliyunSlbListener() *schema.Resource {
 						"retrive_slb_proto": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							Default:  false,
+							Default:  true,
```
YY-tomorrow: I check the slb_listener docs, retrive_slb_proto will be set to false by default. it's XForwardedFor_proto

xuqingfeng: Hmm, that's weird, can you point me which docs you are referring to?

YY-tomorrow: it's CreateLoadBalancerHTTPListener, the retrive_slb_proto is correspond XForwardedFor_proto

xuqingfeng: Can you show me the direct link of the code or docs? 

Because I can see here https://github.com/terraform-providers/terraform-provider-alicloud/blob/master/website/docs/r/slb_listener.html.markdown#block-x_forwarded_for

屏幕快照 2019-06-12 上午10 20 51

It's `true` by default.

YY-tomorrow: thank you,This is my mistake,I will update the docs

PR #1269, opened 2019-06-12
Improve fc_service use common method

YY-tomorrow:


alicloud/resource_alicloud_fc_service.go
```diff
@@ -301,14 +292,14 @@ func parseVpcConfig(d *schema.ResourceData, meta interface{}) (config *fc.VPCCon
 			return
 		}
 		if role, ok := d.GetOk("role"); !ok || role.(string) == "" {
-			err = fmt.Errorf("'role' is required when 'vpc_config' is set.")
+			err = Error("'role' is required when 'vpc_config' is set.")
```
ddaddaddda: The `WrapError` is not used on the line 290

PR #1399, opened 2019-07-05
fix(oss): fixed timeout error in ap-south-1.

Lexsss:


xiaozhu36:
duplicated with #1400

PR #1406, opened 2019-07-10
improve(kms_key): modify the param `description` to forcenew

Lexsss:
…cases.

### DIFF
--- a/alicloud/resource_alicloud_kms_key.go
+++ b/alicloud/resource_alicloud_kms_key.go
@@ -2,11 +2,8 @@ package alicloud
 
 import (
 	"fmt"
-	"time"
-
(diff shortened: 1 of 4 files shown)


alicloud/resource_alicloud_kms_key.go
```diff
@@ -77,8 +75,8 @@ func resourceAlicloudKmsKeyCreate(d *schema.ResourceData, meta interface{}) erro
 	if err != nil {
 		return WrapErrorf(err, DefaultErrorMsg, "kms_key", request.GetActionName(), AlibabaCloudSdkGoERROR)
```
YY-tomorrow: please use alicloud_kms_key, and add debug

PR #1415, opened 2019-07-12
docs(format): Convert inline HCL configs to canonical format.

imjoey:
Along with the new Terraform v0.12, Terraform also introduces the recommended [Idiomatic Style Conventions](https://www.terraform.io/docs/configuration/style.html) for HCL languages. Besides the HCL explicitly defined in `.tf` and . `.tfvars`, the inline HCLs in docs should also follow that rule. This PR is trying to convert all the existing inline HCL codes into the canonical format. The overall formatting process is done by a tool I just developed, I've reviewed the changes twice. Still sorry for the heavy workload on reviewing this PR for me. Thanks.

Signed-off-by: imjoey <majunjiev@gmail.com>

PR #1423, opened 2019-07-14
improve(docs): Reformat abnormal inline HCL code in docs

imjoey:
The previous PR-#1415 formatted plenty of HCL codes in docs. With the release of v1.52.0, I found that some of inline HCL codes were not formatted yet. 

After some digs on it, I figured it out that the tool I built has two limitations, which may cause the codes failed formatted:
1. It works only when the HCL code are valid; (This is a feature to avoid meaningless formatting.)
2. Due to a upstream issue https://github.com/russross/blackfriday/issues/423 , my tool cannot precisely parse the markdown file with CRLF line endings (mostly in Windows format);

Changes proposed in this PR to resolve the failures caused by the limitations above:

* Set `\n` as the line ending in `r/nas_mount_targets.html.markdown` and `r/oss_bucket.html.markdown` instead of `\r\n`
* Fix the code in `r/network_interface_attachment.html.markdown`, `r/pvtz_zone_record.html.markdown`, `r/ram_group_membership.html.markdown` and `r/slb_acl.html.markdown` to in valid HCL format
* Format the newly added `d/ons_instances.html.markdown` after last formatting

... And then format the inline HCL code as expected in the docs above.

Signed-off-by: imjoey <majunjiev@gmail.com>

### DIFF
--- a/website/docs/r/network_interface_attachment.html.markdown
+++ b/website/docs/r/network_interface_attachment.html.markdown
@@ -17,7 +17,6 @@ For information about Elastic Network Interface and how to use it, see [Elastic 
 Bacis Usage
 
 ```
-...
 variable "name" {
   default = "networkInterfaceAttachment"
 }
@@ -27,7 +26,7 @@ variable "number" {
 }
 
 resource "alicloud_vpc" "vpc" {
-  name = "${var.name}"
+  name       = "${var.name}"
   cidr_block = "192.168.0.0/24"
 }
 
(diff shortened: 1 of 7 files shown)


website/docs/r/network_interface_attachment.html.markdown
```diff
@@ -59,31 +59,31 @@ data "alicloud_images" "default" {
...
 resource "alicloud_network_interface_attachment" "attachment" {
-  count = "${var.number}"
-  instance_id = "${element(alicloud_instance.instance.*.id, count.index)}"
+  count                = "${var.number}"
+  instance_id          = "${element(alicloud_instance.instance.*.id, count.index)}"
   network_interface_id = "${element(alicloud_network_interface.interface.*.id, count.index)}"
 }
-...
+# ...
 ```
```
xiaozhu36: I think the line 20 and the line 86 are redundant and them should be removed.

imjoey: OK, it's done. Thanks for reviewing.

PR #1429, opened 2019-07-15
improve(drds): Added descriptions to attributes of datasource.

Lexsss:


alicloud/data_source_alicloud_drds_instances.go
```diff
@@ -29,6 +29,11 @@ func dataSourceAlicloudDRDSInstances() *schema.Resource {
 				MinItems: 1,
```
xiaozhu36: I think the MinItems should be removed.

website/docs/d/drds_instances.html.markdown
```diff
@@ -35,6 +35,8 @@ The following arguments are supported:
 ## Attributes Reference
 
 The following attributes are exported in addition to the arguments listed above:
+
+ * `names` - A list of DRDS descriptions. 
  * `ids` - A list of DRDS instance IDs.
```
xiaozhu36: The line 40 should be `instances`

PR #1439, opened 2019-07-16
improve(mns_queues): Added names to datasource attributes.

Lexsss:


alicloud/data_source_alicloud_mns_queues.go
```diff
@@ -16,12 +14,22 @@ func dataSourceAlicloudMNSQueues() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"ids": {
```
xiaozhu36: There is no need to add ids because of its values equal to names

PR #1492, opened 2019-07-25
Improve(mns_topic_subscription): Improve its resource use common method.

Lexsss:


alicloud/service_alicloud_mns.go
```diff
@@ -106,3 +101,49 @@ func (s *MnsService) WaitForMnsTopic(id string, status Status, timeout int) erro
...
+		if err != nil {
+			if NotFoundError(err) {
+				if status == Deleted {
+					return nil
+				}
+			} else {
+				return WrapError(err)
+			}
+		}
+		if object.TopicName+":"+object.SubscriptionName == id && status != Deleted {
```
xiaozhu36: object may be nil

PR #1514, opened 2019-07-30
support filter invalid name for namespace

wangwenpei:
The aliyun docker registry service is not allowed the below value as namespace:
public, share, primary, everyone, private

I think we should limit it in terraform, instead return API errors.

PS: this is my first time to contribute a Go project, so please check it carefully. I am not familiar for Go.


Lexsss:
Thank you for support our project, please make your code pass the CI check, you can run `make fmt && make test && make vendor-status && make vet`

wangwenpei:
@Lexsss sorry for late, I am too busy these days.
please check it.

PR #1589, opened 2019-08-22
add support to Ipsec_pfs field be set with "disabled" and add example files

zhengjc2018:


alicloud/resource_alicloud_vpn_connection.go
```diff
@@ -153,7 +153,7 @@ func resourceAliyunVpnConnection() *schema.Resource {
 							Type:         schema.TypeString,
 							Optional:     true,
 							Default:      VPN_PFS_G2,
-							ValidateFunc: validateAllowedStringValue([]string{VPN_PFS_G1, VPN_PFS_G2, VPN_PFS_G5, VPN_PFS_G14, VPN_PFS_G24}),
+							ValidateFunc: validateAllowedStringValue([]string{VPN_PFS_G1, VPN_PFS_G2, VPN_PFS_G5, VPN_PFS_G14, VPN_PFS_G24, VPN_PFS_DISABLED}),
```
xiaozhu36: Please add the new value in the docs.

alicloud/resource_alicloud_vpn_connection.go
```diff
@@ -134,7 +134,7 @@ func resourceAliyunVpnConnection() *schema.Resource {
 			"ipsec_config": {
 				Type:     schema.TypeList,
 				Optional: true,
-				Computed: true,
+				// Computed: true,
```
xiaozhu36: why remove the line 137?

examples/vpn-connection/main.tf
```diff
@@ -0,0 +1,38 @@
+resource "alicloud_vpc" "default" {
```
xiaozhu36: please run `terraform fmt` before committing the code.

zhengjc2018:
i think that "computed:true" means the value can not be set, only can know
after apply. But actually, it can be set as normal arguments and has its
default value.So i remove the line 137

He Guimin <notifications@github.com> 于2019年8月22日周四 下午8:40写道：

zhengjc2018:
fix fmt error and fix docs error

PR #1602, opened 2019-08-27
fix(rds): add diffsuppressfunc to rds tags.

Lexsss:


alicloud/resource_alicloud_db_instance.go
```diff
@@ -268,7 +268,27 @@ func resourceAlicloudDBInstance() *schema.Resource {
...
+							return false
+						} else {
+							if strings.ToLower(ovv.(string)) != strings.ToLower(nvv.(string)) {
+								return false
+							}
+						}
+					}
+					return true
+				},
+			},
```
xiaozhu36: unit test?
Please add a note to show the point in the docs.

PR #1615, opened 2019-08-29
improve(gpdb): gpdb instance supported tags.

Lexsss:


alicloud/service_alicloud_gpdb.go
```diff
@@ -177,3 +181,73 @@ func (s *GpdbService) WaitForGpdbConnection(id string, status Status, timeout in
 		}
 	}
 }
+
+func (s *GpdbService) setInstanceTags(d *schema.ResourceData) error {
+	if d.HasChange("tags") {
+		oraw, nraw := d.GetChange("tags")
+		o := oraw.(map[string]interface{})
+		n := nraw.(map[string]interface{})
+		create, remove := diffGpdbTags(gpdbTagsFromMap(o), gpdbTagsFromMap(n))
```
xiaozhu36: I think just `d.getChange()` is ok.

PR #1621, opened 2019-08-30
improve(vpc): vpc resources and datasources supported tags.

Lexsss:


alicloud/resource_alicloud_vpc.go
```diff
@@ -180,30 +185,35 @@ func resourceAliyunVpcRead(d *schema.ResourceData, meta interface{}) error {
...
-	request.VpcId = d.Id()
-
-	if d.HasChange("name") {
-		request.VpcName = d.Get("name").(string)
-		attributeUpdate = true
+	vpcService := VpcService{client}
+	if err := vpcService.setInstanceTags(d, TagResourceVpc); err != nil {
+		return WrapError(err)
 	}
+	if !d.IsNewResource() {
```
xiaozhu36:
```
if d.IsResource() {
   d.Partial(false)
   return resourceAliyunVpcRead(d, meta)
}
```

PR #1748, opened 2019-10-15
improve(slb): updated slb rule testcase. 

Threadv:
improve(slb): updated slb rule testcase.

### DIFF
--- a/alicloud/resource_alicloud_slb_rule_test.go
+++ b/alicloud/resource_alicloud_slb_rule_test.go
@@ -6,20 +6,20 @@ import (
 
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/slb"
 	"github.com/hashicorp/terraform/helper/resource"
-	"github.com/hashicorp/terraform/terraform"
 	"github.com/terraform-providers/terraform-provider-alicloud/alicloud/connectivity"
 )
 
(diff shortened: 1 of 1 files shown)


alicloud/resource_alicloud_slb_rule_test.go
```diff
@@ -49,13 +48,35 @@ func TestAccAlicloudSlbRuleUpdate(t *testing.T) {
 				ImportStateVerify: true,
 			},
 			{
-				Config: testAccSlbRuleBasic_server_group_id,
+				Config: testAccConfig(map[string]interface{}{}),
```
Lexsss: No need a step with an empty map.

alicloud/resource_alicloud_slb_rule_test.go
```diff
@@ -49,13 +48,35 @@ func TestAccAlicloudSlbRuleUpdate(t *testing.T) {
...
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheck(nil),
 				),
 			},
 			{
-				Config: testAccSlbRuleBasic_health_check,
+				Config: testAccConfig(map[string]interface{}{
```
Lexsss: The step set all fields should be move to the end of all steps.

PR #1815, opened 2019-10-31
imporve(instance): modified the argument reference in doc.

Lexsss:


website/docs/r/instance.html.markdown
```diff
@@ -118,7 +118,7 @@ On other OSs such as Linux, the host name can contain a maximum of 30 characters
 * `role_name` - (Optional, Force new resource) Instance RAM role name. The name is provided and maintained by RAM. You can use `alicloud_ram_role` to create a new one.
```
xiaozhu36: * `user_data` - (Optional) User-defined data to customize the startup behaviors of an ECS instance and to pass data into an ECS instance. From version 1.60.0, it can be update in-place. If updated, the instance will reboot to make the change take effect. Note: Not all of changes will take effect and it depends on [cloud-init module type](https://cloudinit.readthedocs.io/en/latest/topics/modules.html )

PR #1858, opened 2019-11-13
Add resources and datasources to encrypt and decrypt data with KMS

gonzolino:
This PR adds the resource `alicloud_kms_ciphertext` as well as the two data sources `alicloud_kms_ciphertext` and `alicloud_kms_secret`. They allow to encrypt  secrets and decrypt cipher texts using Alibaba Cloud KMS.

### DIFF
--- a/alicloud/data_source_alicloud_kms_ciphertext.go
+++ b/alicloud/data_source_alicloud_kms_ciphertext.go
@@ -1,0 +1,68 @@
+package alicloud
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/aliyun/alibaba-cloud-sdk-go/services/kms"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-alicloud/alicloud/connectivity"
+)
+
+func dataSourceAlicloudKmsCiphertext() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAlicloudKmsCiphertextRead,
+
+		Schema: map[string]*schema.Schema{
+			"plaintext": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+
+			"key_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"context": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"ciphertext_blob": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAlicloudKmsCiphertextRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*connectivity.AliyunClient)
+
+	// Since a ciphertext has no ID, we create an ID based on
+	// current unix time.
+	d.SetId(strconv.FormatInt(time.Now().Unix(), 16))
+
+	request := kms.CreateEncryptRequest()
+	request.Plaintext = d.Get("plaintext").(string)
+	request.KeyId = d.Get("key_id").(string)
+	request.RegionId = client.RegionId
+
+	if context := d.Get("context"); context != nil {
+		request.EncryptionContext = context.(string)
+	}
+
+	raw, err := client.WithKmsClient(func(kmsClient *kms.Client) (interface{}, error) {
+		return kmsClient.Encrypt(request)
+	})
+	if err != nil {
+		return WrapErrorf(err, DefaultErrorMsg, "alicloud_kms_ciphertext", request.GetActionName(), AlibabaCloudSdkGoERROR)
+	}
+	addDebug(request.GetActionName(), raw, request.RpcRequest, request)
+	response, _ := raw.(*kms.EncryptResponse)
+	d.Set("ciphertext_blob", response.CiphertextBlob)
+
+	return nil
+}
--- a/alicloud/data_source_alicloud_kms_ciphertext.go
+++ b/alicloud/data_source_alicloud_kms_ciphertext.go
@@ -5,7 +5,7 @@ import (
 	"time"
 
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/kms"
-	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-alicloud/alicloud/connectivity"
 )
 
--- a/alicloud/data_source_alicloud_kms_ciphertext.go
+++ b/alicloud/data_source_alicloud_kms_ciphertext.go
@@ -25,9 +25,12 @@ func dataSourceAlicloudKmsCiphertext() *schema.Resource {
 				Required: true,
 			},
 
-			"context": {
-				Type:     schema.TypeString,
+			"encryption_context": {
+				Type:     schema.TypeMap,
 				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
 			},
 
 			"ciphertext_blob": {
@@ -50,8 +53,13 @@ func dataSourceAlicloudKmsCiphertextRead(d *schema.ResourceData, meta interface{
 	request.KeyId = d.Get("key_id").(string)
 	request.RegionId = client.RegionId
 
-	if context := d.Get("context"); context != nil {
-		request.EncryptionContext = context.(string)
+	if context := d.Get("encryption_context"); context != nil {
+		cm := context.(map[string]interface{})
+		contextJson, err := convertMaptoJsonString(cm)
+		if err != nil {
+			return WrapErrorf(err, DefaultErrorMsg, "alicloud_kms_ciphertext", request.GetActionName(), AlibabaCloudSdkGoERROR)
+		}
+		request.EncryptionContext = string(contextJson)
 	}
 
 	raw, err := client.WithKmsClient(func(kmsClient *kms.Client) (interface{}, error) {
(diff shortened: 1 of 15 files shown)


gonzolino:
Hi @xiaozhu36, It would be nice if you could have a look at my PR.

alicloud/data_source_alicloud_kms_ciphertext.go
```diff
@@ -0,0 +1,68 @@
...
+				Required:  true,
+				Sensitive: true,
+			},
+
+			"key_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"context": {
```
xiaozhu36: I think use `encryption_context ` will be better. 
In addition, it should be `TypeMap`.

alicloud/data_source_alicloud_kms_ciphertext_test.go
```diff
@@ -0,0 +1,46 @@
...
+						resourceId, "ciphertext_blob",
+					),
+				),
+			},
+			{
+				ResourceName:      resourceId,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
```
xiaozhu36: There is no need to test `import`, so the line 28-32 should be removed.

alicloud/data_source_alicloud_kms_secret.go
```diff
@@ -0,0 +1,67 @@
...
+			},
+
+			"key_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"context": {
+				Type:     schema.TypeString,
+				Optional: true,
```
xiaozhu36: still applies.

alicloud/data_source_alicloud_kms_secret_test.go
```diff
@@ -0,0 +1,53 @@
...
+					resource.TestCheckResourceAttrSet(
+						resourceId, "key_id",
+					),
+				),
+			},
+			{
+				ResourceName:      resourceId,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
xiaozhu36: still applies.

alicloud/provider.go
```diff
@@ -128,6 +128,8 @@ func Provider() terraform.ResourceProvider {
 			"alicloud_eips":                   dataSourceAlicloudEips(),
 			"alicloud_key_pairs":              dataSourceAlicloudKeyPairs(),
 			"alicloud_kms_keys":               dataSourceAlicloudKmsKeys(),
+			"alicloud_kms_ciphertext":         dataSourceAlicloudKmsCiphertext(),
+			"alicloud_kms_secret":             dataSourceAlicloudKmsSecret(),
```
xiaozhu36: I think use alicloud_kms_plaintext will be better.

alicloud/provider.go
```diff
@@ -305,6 +307,7 @@ func Provider() terraform.ResourceProvider {
 			"alicloud_key_pair":                      resourceAlicloudKeyPair(),
 			"alicloud_key_pair_attachment":           resourceAlicloudKeyPairAttachment(),
 			"alicloud_kms_key":                       resourceAlicloudKmsKey(),
+			"alicloud_kms_ciphertext":                resourceAlicloudKmsCiphertext(),
```
xiaozhu36: I think there is no need to add a resource to get ciphertext

alicloud/data_source_alicloud_kms_ciphertext.go
gonzolino: Agreed, I changed the parameter accordingly.

alicloud/data_source_alicloud_kms_ciphertext_test.go
gonzolino: Removed

alicloud/data_source_alicloud_kms_secret_test.go
gonzolino: done

alicloud/provider.go
gonzolino: Ok, I renamed `alicloud_kms_secret` to `alicloud_kms_plaintext`

alicloud/provider.go
gonzolino: The idea behind the ciphertext resource is that it stays stable across applies.
A data source re-encrypts the data on each apply, because it runs the Read method on each apply. A resource will only encrypt data when the parameters change.

So if you want to use the data source ciphertext in another resource (e.g. inside the user-data for a VM), terraform would replace that resource on every apply because the ciphertexts changes. If you would use the ciphertext resource in such a case, terraform would only replace the resource if a re-encryption is necessary.

alicloud/data_source_alicloud_kms_secret.go
gonzolino: done

website/docs/d/kms_ciphertext.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `plaintext` - The plaintext to be encrypted which must be encoded in Base64.
+* `key_id` - The globally unique ID of the CMK.
+* `encryption_context` -
+  (Optional) The Encryption context. If you specify this parameter here, it is also required when you call the Decrypt API operation. For more information, see [Encryption Context](https://www.alibabacloud.com/help/doc-detail/42975.htm?spm=a2c63.p38356.b99.14.47562193BvC7Hu).
+
```
xiaozhu36: Please remove the needless string `?spm=a2c63.p38356.b99.14.47562193BvC7Hu` in the line 36.

website/docs/d/kms_secret.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+  value = data.alicloud_kms_secret.secret.plaintext
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `encryption_context` -
+  (Optional) The Encryption context. If you specify this parameter in the Encrypt or GenerateDataKey API operation, it is also required when you call the Decrypt API operation. For more information, see [Encryption Context](https://www.alibabacloud.com/help/doc-detail/42975.htm?spm=a2c63.p38356.b99.14.47562193BvC7Hu).
```
xiaozhu36: still applies

website/docs/r/kms_ciphertext.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `plaintext` - (ForceNew) The plaintext to be encrypted which must be encoded in Base64.
+* `key_id` - (ForceNew) The globally unique ID of the CMK.
+* `encryption_context` -
+  (Optional, ForceNew) The Encryption context. If you specify this parameter here, it is also required when you call the Decrypt API operation. For more information, see [Encryption Context](https://www.alibabacloud.com/help/doc-detail/42975.htm?spm=a2c63.p38356.b99.14.47562193BvC7Hu).
```
xiaozhu36: still applies

alicloud/provider.go
xiaozhu36: I think, for data source, if keyId and plaintext are not changed, the ciphertext will not be changed. Right?

website/docs/d/kms_ciphertext.html.markdown
gonzolino: fixed

gonzolino: fixed

website/docs/d/kms_secret.html.markdown
gonzolino: fixed

website/docs/r/kms_ciphertext.html.markdown
gonzolino: fixed

alicloud/provider.go
gonzolino: AFAIK the `Read` method of a data source is executed on every apply. So the plaintext will be re-encrypted on every apply. The encryption mechanism used by KMS is randomized (for increased security), so each generated ciphertext will be different, even if the key and plaintext stay the same.

So in effect terraform will detect a changed ciphertext on each apply.

xiaozhu36:
HI @gonzolino Please rebase this PR code with master and merge all of commit.

gonzolino:
@xiaozhu36 Rebased on master :)

website/docs/d/kms_ciphertext.html.markdown
```diff
@@ -0,0 +1,42 @@
+---
+layout: "alicloud"
```
xiaozhu36: please add a subcategory in the line 2: subcategory: "KMS", like other docs.

website/docs/d/kms_plaintext.html.markdown
```diff
@@ -0,0 +1,53 @@
+---
+layout: "alicloud"
```
xiaozhu36: still applies

website/docs/r/kms_ciphertext.html.markdown
```diff
@@ -0,0 +1,43 @@
+---
```
xiaozhu36: still applies

website/docs/d/kms_ciphertext.html.markdown
gonzolino: done

website/docs/d/kms_plaintext.html.markdown
gonzolino: done

website/docs/r/kms_ciphertext.html.markdown
gonzolino: done

gonzolino:
Rebased on latest master and updated dependencies to `github.com/hashicorp/terraform-plugin-sdk/xxx`.

PR #1885, opened 2019-11-19
Do not assume `private_zone` is returned from API

### LINKED ISSUE #1866: alicloud_cs_serverless_kubernetes recreates cluster if private_zone=true
### Terraform Version
terraform: 0.12.14
alicloud provider: 1.62

### Region
us-west-1

### Affected Resource(s)
- alicloud_cs_serverless_kubernetes

### Terraform Configuration Files
```hcl
resource "alicloud_cs_serverless_kubernetes" "serverless" {
  name_prefix                    = "alias-k8s-test"
  vpc_id                         = alicloud_vpc.vpc.id
  vswitch_id                     = alicloud_vswitch.vs.id
  new_nat_gateway                = false
  endpoint_public_access_enabled = false
  private_zone                   = true
  force_update                   = true
  deletion_protection            = false
  kube_config                    = "/tmp/.kube/config"
  client_cert                    = "/tmp/.kube/client-cert.pem"
  client_key                     = "/tmp/.kube/client-key.pem"
  cluster_ca_cert                = "/tmp/.kube/cluster-ca-cert.pem"
  provider                       = alicloud.argon-ali-terraform-us
}
```

### Debug Output
```
2019/11/15 00:39:08 [WARN] Provider "alicloud" produced an unexpected new value for alicloud_cs_serverless_kubernetes.serverless, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .private_zone: was cty.True, but now cty.False
```

### Expected Behavior
The cluster should remain untouched if no configuration changes happen

### Actual Behavior
The cluster is recreated every time terraform plan/apply is executed

### Steps to Reproduce
1. `terraform apply`

leonrodenburg:
Container Service endpoint `/clusters` does not return information about whether or not a private zone was configured for the cluster. Currently, if a cluster is created in Terraform with the `private_zone` property set to `true`, every deploy will cause a new cluster to be created as Golang booleans default to `false` and that property has `ForceNew` set to `true`.

Resolves #1866

### DIFF
--- a/alicloud/data_source_alicloud_cs_serverless_kubernetes_clusters.go
+++ b/alicloud/data_source_alicloud_cs_serverless_kubernetes_clusters.go
@@ -60,10 +60,6 @@ func dataSourceAlicloudCSServerlessKubernetesClusters() *schema.Resource {
 							Type:     schema.TypeBool,
 							Computed: true,
 						},
-						"private_zone": {
-							Type:     schema.TypeBool,
-							Computed: true,
-						},
 						"deletion_protection": {
 							Type:     schema.TypeBool,
 							Computed: true,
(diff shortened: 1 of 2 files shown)


menglingwei:
@xiaozhu36

PR #1888, opened 2019-11-21
improve(provider): use common method to instead of several duplicate methods

xiaozhu36:
…


xiaozhu36:
Duplicated with #1912

PR #1922, opened 2019-11-28
fix(alicloud_cs_serverless_kubernetes): write correct data into kube_config

chanind:
fixes https://github.com/terraform-providers/terraform-provider-alicloud/issues/1867

This PR fixes a bug where `alicloud_cs_serverless_kubernetes` resource writes an empty string into the file passed in to `kube_config`. This bug makes it impossible to use the `alicloud_cs_serverless_kubernetes` resource in terraform since there's no way to pass connection info to other terraform resources after creation.

I'm not a go programmer, and I don't really understand why re-encoding the JSON response from alicloud and then re-parsing it fixes the bug. If there's a more direct way to fix this let me know and I'll update the PR (or feel free to update the PR directly).


chanind:
Closing since it looks like #1923 does the same thing

PR #1941, opened 2019-12-04
feat(market): added market products datasource.

Lexsss:


website/alicloud.erb
```diff
@@ -783,6 +783,20 @@
                   </ul>
                 </li>
 
+                <li>
+                  <a href="#">Market</a>
```
xiaozhu36: please use `Market Place` instead `Market`

website/docs/d/market_products.html.markdown
```diff
@@ -0,0 +1,64 @@
+---
+subcategory: "MARKET"
```
xiaozhu36: I think it should be Market Place

website/docs/index.html.markdown
```diff
@@ -265,6 +265,8 @@ Nested `endpoints` block supports the following:
 
 * `ddoscoo` - (Optional) Use this to override the default endpoint URL constructed from the `region`. It's typically used to connect to custom BGP-Line Anti-DDoS Pro endpoints.
 
+* `market` - (Optional) Use this to override the default endpoint URL constructed from the `region`. It's typically used to connect to custom Market endpoints.
```
xiaozhu36: please add endpoint in the `client.go`

PR #1979, opened 2019-12-13
add output to ssl vpn client cert

ionTea:
Added additional outputs to the `alicloud_ssl_vpn_client_cert` resource. They were already available on the API result, but weren't passed to the terraform resource outputs.

### DIFF
--- a/alicloud/resource_alicloud_ssl_vpn_client_cert.go
+++ b/alicloud/resource_alicloud_ssl_vpn_client_cert.go
@@ -37,6 +37,26 @@ func resourceAliyunSslVpnClientCert() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"ca_cert": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+			"client_cert": {
+				Type:      schema.TypeString,
+				Computed:  true,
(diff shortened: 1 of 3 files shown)


xiaozhu36:
HI . @ionTea Thanks for your contribution. Can you add the fields to test case?

ionTea:
Hello, a previous bug caused these tests to be broken already. Luckily https://github.com/terraform-providers/terraform-provider-alicloud/pull/1977 fixed the issue.

Please see the new updated tests for the new fields.

alicloud/resource_alicloud_ssl_vpn_client_cert.go
```diff
@@ -37,6 +37,26 @@ func resourceAliyunSslVpnClientCert() *schema.Resource {
...
+			"client_key": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+			"client_config": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
```
xiaozhu36: Please add docs for above four new parameters.

ionTea: How do I add docs? Do I just manually add it, or are they generated somehow?

PR #2013, opened 2019-12-24
Update go sdk for ons

Rich-yi:


xiaozhu36:
Duplicated with #2014

PR #2016, opened 2019-12-24
fix:resource_alicloud_db_instance(modify spec & change pay type)

zhaiyongsheng0123:
=== RUN   TestAccAlicloudDBInstanceMysql
--- PASS: TestAccAlicloudDBInstanceMysql (4572.83s)
PASS

### DIFF
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -52,6 +52,11 @@ const (
 )
 
 const (
+	Down = "Down"
+	Up   = "Up"
+)
+
+const (
 	NormalMode = "normal"
 	SafetyMode = "safety"
 )
--- a/alicloud/resource_alicloud_db_instance.go
+++ b/alicloud/resource_alicloud_db_instance.go
@@ -58,6 +58,13 @@ func resourceAlicloudDBInstance() *schema.Resource {
 				Required: true,
 			},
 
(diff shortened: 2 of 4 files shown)


alicloud/resource_alicloud_db_instance.go
```diff
@@ -58,6 +58,13 @@ func resourceAlicloudDBInstance() *schema.Resource {
 				Required: true,
 			},
 
+			"direction": {
+				Type:         schema.TypeString,
+				ValidateFunc: validation.StringInSlice([]string{Up, Down}, false),
```
Lexsss: specify string directly

alicloud/common.go
```diff
@@ -51,6 +51,11 @@ const (
 	Postpaid = PayType("Postpaid")
 )
 
+const (
```
Lexsss: no need const

alicloud/resource_alicloud_db_instance.go
```diff
@@ -58,6 +58,13 @@ func resourceAlicloudDBInstance() *schema.Resource {
 				Required: true,
 			},
 
+			"direction": {
```
Lexsss: does the response of describe api has this field?

PR #2029, opened 2019-12-27
fix(alikafka) the region has duplicateName.

zql-tqy:


alicloud/resource_alicloud_alikafka_instance_test.go
```diff
@@ -92,7 +92,7 @@ func TestAccAlicloudAlikafkaInstance_basic(t *testing.T) {
 
 	rand := acctest.RandInt()
 	testAccCheck := rac.resourceAttrMapUpdateSet()
-	name := fmt.Sprintf("tf-testacc-alikafkainstancebasic%v", rand)
+	name := fmt.Sprintf("tf-testacc-%salikafkainstancebasic%v", defaultRegionToTest, rand)
```
Lexsss: modify sweeper and check name

PR #2037, opened 2019-12-30
Use public endpoint for Kube config if public access enabled

leonrodenburg:
When reading a serverless K8s cluster, there was a boolean logic error that caused private IP addresses to be set in the Kube config file when public API access is enabled. This PR flips the logic so a public IP is set in the Kube config YAML when public access is enabled, and uses a private IP if public access is disabled.


leonrodenburg:
@xiaozhu36 Any chance this could get merged? Really need this :)

PR #2054, opened 2020-01-07
int to string

juhuahaoteng:
image_map["size"] int to string


juhuahaoteng:
image_map["size"] int to string

xiaozhu36:
Hi, I think this pr duplicates with #2052 , right?

juhuahaoteng:
sorry

PR #2057, opened 2020-01-07
adding RAM role_name to data.alicloud_instances

chanind:
This PR adds `role_name` to the output for the `alicloud_instances` data source. This mirrors what's output for the `alicloud_instance` resource. 

Closes https://github.com/terraform-providers/terraform-provider-alicloud/issues/2053


xiaozhu36:
HI . @chanind Thanks for your contribution. But, I think this PR has low efficiency when there are lots of instance. And I have improved it by #2060 and it supports fetch instances  by `ram_role_name`.

chanind:
Awesome, that's even better! Thanks @xiaozhu36!

PR #2065, opened 2020-01-08
fix vpn bug

zhengjc2018:


alicloud/resource_alicloud_vpn_gateway.go
```diff
@@ -29,10 +29,10 @@ func resourceAliyunVpnGateway() *schema.Resource {
...
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				ValidateFunc: validation.StringLenBetween(1, 128),
-				Default:      resource.PrefixedUniqueId("tf-vpn-"),
+				Type:     schema.TypeString,
+				Optional: true,
+				//ValidateFunc: validation.StringLenBetween(1, 128),
+				Default: "tf-vpn",
```
xiaozhu36: I think this default should not be modified.

zhengjc2018:
first, the name is optional and if not be given, the vpn will be created
with the name by it's VpnGatewayId. In the latest commit, i remove the
default and set `computed: true`
secondly, it no need to be unique, users can create vpns with the same name.
last, if use `resource.PrefixedUniqueId("tf-vpn-")`, it may cause error
[image: image.png]
[image: image.png]

He Guimin <notifications@github.com> 于2020年1月15日周三 下午3:51写道：

PR #2083, opened 2020-01-10
improve(polardb): fix polardb add parameters bug

zhangyuangang:


alicloud/resource_alicloud_polardb_cluster.go
```diff
@@ -161,7 +161,7 @@ func resourceAlicloudPolarDBClusterUpdate(d *schema.ResourceData, meta interface
 	d.Partial(true)
 
 	if d.HasChange("parameters") {
-		if err := polarDBService.ModifyParameters(d, d.Get("parameters").(string)); err != nil {
+		if err := polarDBService.ModifyParameters(d, "parameters"); err != nil {
```
xiaozhu36: I think just use `if err := polarDBService.ModifyParameters(d); err != nil {` is ok.

PR #2171, opened 2020-02-14
add market product image id

956081712:


alicloud/data_source_alicloud_market_product.go
```diff
@@ -64,6 +64,22 @@ func dataSourceAlicloudProduct() *schema.Resource {
 											},
 										},
 									},
+									"image_ids": {
```
xiaozhu36: please change it to `images`

PR #2190, opened 2020-02-21
updated to GO SDK 1.6.387

arafato:
upgraded to GO SDK 1.6.387 (for a new resource provider) and added some now neccessary casts from int64 to int.


alicloud/service_alicloud_pvtz.go
```diff
@@ -136,7 +136,7 @@ func (s *PvtzService) DescribePvtzZoneRecord(id string) (record pvtz.Record, err
 		}
 
 		for _, rec := range response.Records.Record {
-			if strconv.Itoa(rec.RecordId) == parts[0] {
+			if strconv.Itoa(int(rec.RecordId)) == parts[0] {
```
xiaozhu36: Please use safety method `strconv.FormatInt` instead of `int()`

PR #2277, opened 2020-03-19
support parameter DesiredCapacity

juhuahaoteng:
support parameter DesiredCapacity

### DIFF
--- a/alicloud/resource_alicloud_ess_scalinggroup.go
+++ b/alicloud/resource_alicloud_ess_scalinggroup.go
@@ -37,6 +37,11 @@ func resourceAlicloudEssScalingGroup() *schema.Resource {
 				Required:     true,
 				ValidateFunc: validation.IntBetween(0, 1000),
 			},
+			"desired_capacity": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ValidateFunc: validation.IntBetween(0, 1000),
+			},
 			"scaling_group_name": {
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


alicloud/resource_alicloud_ess_scalinggroup.go
```diff
@@ -331,6 +341,9 @@ func buildAlicloudEssScalingGroupArgs(d *schema.ResourceData, meta interface{})
 	if v, ok := d.GetOk("scaling_group_name"); ok && v.(string) != "" {
 		request.ScalingGroupName = v.(string)
 	}
+	if v, ok := d.GetOk("desired_capacity"); ok && v.(int) != 0 {
```
Lexsss: it can't be 0?

website/docs/r/ess_scaling_group.html.markdown
```diff
@@ -95,6 +95,7 @@ The following arguments are supported:
 
 * `min_size` - (Required) Minimum number of ECS instances in the scaling group. Value range: [0, 1000].
 * `max_size` - (Required) Maximum number of ECS instances in the scaling group. Value range: [0, 1000].
+* `desired_capacity` - (Optional) Expected number of ECS instances in the scaling group. Value range: [min_size, max_size].
```
Lexsss: (Optional, Available in 1.76.0+), and please ensure the description is right.

alicloud/resource_alicloud_ess_scalinggroup.go
juhuahaoteng: no, it's desired value,the min value is 1.

PR #2304, opened 2020-03-25
delete vswitch timeout

renorzr:
延长vswitch删除超时时间


alicloud/resource_alicloud_vswitch.go
```diff
@@ -161,7 +161,7 @@ func resourceAliyunSwitchDelete(d *schema.ResourceData, meta interface{}) error
 	request := vpc.CreateDeleteVSwitchRequest()
 	request.RegionId = client.RegionId
 	request.VSwitchId = d.Id()
-	err := resource.Retry(6*time.Minute, func() *resource.RetryError {
+	err := resource.Retry(20*time.Minute, func() *resource.RetryError {
```
xiaozhu36, line 164: I think it should use `timeout` to support self-define.

xiaozhu36 (changes requested):

PR #2408, opened 2020-04-30
Let WriteFile return write/delete errors

### LINKED ISSUE #2406: alicloud_cs_managed_kubernetes won't error if K8s config cannot be written
If the `alicloud_cs_managed_kubernetes` resource fails to write the K8s config file, it actually won't report any error at the moment, which causes trouble if you then rely on that file to exist in a CI pipeline, etc.

One situation when the writing might fail is in case the folder doesn't exist, e.g.

```
resource "alicloud_cs_managed_kubernetes" "k8s" {
  kube_config = "/some/nonexisting/dir/config"
  ...
}
```

The resource should make that error explicit instead of hiding it from the user.

Provider version: 1.80.1

mlafeldt:
Fixes #2406

### DIFF
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -454,7 +454,9 @@ func writeToFile(filePath string, data interface{}) error {
 		}
 	}
 
-	os.Remove(filePath)
+	if err := os.Remove(filePath); err != nil {
+		return err
+	}
 
 	var out string
 	switch data.(type) {
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -471,8 +471,7 @@ func writeToFile(filePath string, data interface{}) error {
 		out = string(bs)
 	}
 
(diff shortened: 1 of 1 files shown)


alicloud/common.go
```diff
@@ -454,7 +454,9 @@ func writeToFile(filePath string, data interface{}) error {
 		}
 	}
 
-	os.Remove(filePath)
+	if err := os.Remove(filePath); err != nil {
+		return err
+	}
```
mlafeldt, line 459: I'd actually prefer not doing `os.Remove` at all, since `ioutil.WriteFile` does most of the work already:

```
WriteFile writes data to a file named by filename. If the file does not exist, WriteFile creates it with permissions perm (before umask); otherwise WriteFile truncates it before writing.
```

However, I'm not sure if this will have any side effects since the function is widely used.

PR #2410, opened 2020-04-30
fix sls log regionid

xuxiaoahang2018:


xiaozhu36:
Duplicated with #2411

PR #2412, opened 2020-04-30
Fix regression in writeToFile

mlafeldt:
Fixes a regression introduced by #2408 and makes the function less destructive.


mlafeldt:
@xiaozhu36 PTAL

PR #2420, opened 2020-05-07
export alicloud_cs_serverless_kubernetes client_cert/client_key/cluster_ca_cert/kube_config as attributes so we can pass them directly to kubernetes provider

lonegunmanb:
One common scenario of using alicloud_cs_serverless_kubernetes is we create a cluster and store it's key information in a backend, then another terraform stack can read these keys from remote state then provision the cluster. If we want to do so now, we need write these key informations into four files and read them into backend. Imagine if you'are using CI, and your CI agent is a container in a K8s cluster, the runner who created cluster is different than the runner who need to destroy it, and runner who want to destroy this cluster will find lacking of these four files and throw an error.

So the best way is export these keys directly as attributes so we can store them into backend or pass them to kubernetes provider instead of writing to any file.

### DIFF
--- a/alicloud/resource_alicloud_cs_serverless_kubernetes.go
+++ b/alicloud/resource_alicloud_cs_serverless_kubernetes.go
@@ -99,13 +99,29 @@ func resourceAlicloudCSServerlessKubernetes() *schema.Resource {
 				Type:     schema.TypeMap,
 				Optional: true,
 				ForceNew: true,
-				//ValidateFunc: validateCSClusterTags,
+				// ValidateFunc: validateCSClusterTags,
 			},
 			"force_update": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				ForceNew: true,
 				Default:  false,
+			},
+			"client_cert_content": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"client_key_content": {
+				Type:    schema.TypeString,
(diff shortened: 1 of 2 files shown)


chanind:
👍 was just going to request this same feature! It's really awkward and error-prone to use the current system of writing out configs to files during runs. We need to immediately read-in the files again later in the Terraform run, and terraform doesn't know that the content of these files depends on the output of the `alicloud_cs_serverless_kubernetes` resource. Terraform also freaks out if these files don't exist before the run starts, so you need to manually `touch` the files first. This PR would make all this much much better! Would also love to see this same output in the `alicloud_cs_managed_kubernetes` resource too.

lonegunmanb:
> 👍 was just going to request this same feature! It's really awkward and error-prone to use the current system of writing out configs to files during runs. We need to immediately read-in the files again later in the Terraform run, and terraform doesn't know that the content of these files depends on the output of the `alicloud_cs_serverless_kubernetes` resource. Terraform also freaks out if these files don't exist before the run starts, so you need to manually `touch` the files first. This PR would make all this much much better! Would also love to see this same output in the `alicloud_cs_managed_kubernetes` resource too.

You're right man, I'll submit another pr for alicloud_cs_managed_kubernetes when I got time. But I doubt that they'll accept this pr quickly or not.

PR #2435, opened 2020-05-09
add a new resource alicloud_cen_route_service

SanfordLuo:
add a new resource alicloud_cen_route_service


xiaozhu36:
Hi @SanfordLuo It has been supported by version 1.99.0 and docs in https://registry.terraform.io/providers/aliyun/alicloud/latest/docs/resources/cen_route_service. Please check it.

PR #2464, opened 2020-05-20
Control the instance start and stop through the status attribute.

zql-tqy:


alicloud/resource_alicloud_instance.go
```diff
@@ -651,13 +653,18 @@ func resourceAliyunInstanceUpdate(d *schema.ResourceData, meta interface{}) erro
...
+		target = d.Get("status").(string)
+	}
+	if imageUpdate || vpcUpdate || passwordUpdate || typeUpdate || statusUpdate {
 		run = true
 		instance, errDesc := ecsService.DescribeInstance(d.Id())
 		if errDesc != nil {
 			return WrapError(errDesc)
 		}
-		if instance.Status == string(Running) {
+		if instance.Status == string(Running) || (instance.Status != target && target == string(Stopped)) {
```
xiaozhu36, line 667: I think the line 667 should be `if (statusUpdate && target == string(Stopped)) || instance.Status == string(Running) {`

alicloud/resource_alicloud_instance.go
```diff
@@ -651,13 +653,18 @@ func resourceAliyunInstanceUpdate(d *schema.ResourceData, meta interface{}) erro
 	if err != nil {
 		return WrapError(err)
 	}
-	if imageUpdate || vpcUpdate || passwordUpdate || typeUpdate {
+	target := ""
```
xiaozhu36, line 656: I think the line 656 should be `target = d.Get("status").(string)`

PR #2479, opened 2020-05-25
fix_network_interfaces

juhuahaoteng:
fix_network_interfaces


alicloud/data_source_alicloud_network_interfaces.go
```diff
@@ -65,6 +65,7 @@ func dataSourceAlicloudNetworkInterfaces() *schema.Resource {
 				Type:     schema.TypeList,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Computed: true,
+				Optional: true,
```
xiaozhu36, line 68: There is no need to set `names` to an input parameter and can use `name_regex` instead.

PR #2610, opened 2020-06-30
feat/oss/noncurrent version rule

Akash-Jain-GMI:
Ability to specify lifecycle rules for non current version files OSS buckets


Akash-Jain-GMI:
Approving locally

PR #2670, opened 2020-07-24
Add new product oos template, include resource and datasource

lixue323:


website/docs/d/oos_templates.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+output "first_template_name" {
+  value = "${data.alicloud_oos_templates.example.templates.0.template_name}"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `ids` - (Optional) A list of OOS Template ids. The value is same as template_name.
```
zql-tqy, line 41: Each element in the list is same as template_name.

PR #2673, opened 2020-07-24
fix(edas): improve sweeper test

xiaozhu36:


kmoe:
We don't recommend that upx be used to compress provider binaries, as upx can sometimes be flagged as malware by antivirus.

The latest `terraform-provider-alicloud` release, v1.91.0, is 172MB unzipped, which is comparable with other large providers. Our build process adds the "-s -w" ldflags, which may explain the discrepancy with the binary size you obtained by compiling locally.

The best way to reduce the size of the compiled provider is to remove unnecessary dependencies. One of these is Terraform Core, as described in https://github.com/terraform-providers/terraform-provider-alicloud/issues/2674.

davidji99:
Echoing @kmoe, I would stay away from `upx`. I used `upx` once to compress a CLI tool I wrote for our company and the company antivirus software removed it along with other unrelated binaries. It was a not a great experience for the end users.

xiaozhu36:
Thanks a lot. @kmoe

PR #2697, opened 2020-08-10
transfer terraform-provider-alicloud to aliyun from terraform-providers

xiaozhu36:


bittopaz:
Hey @xiaozhu36, is there any reason to do this transfer, please?

Will the future release cycle be impacted by this?

JanWaller:
When / how are these newly updated providers (e.g., 1.93) available in the terraform registry?

Issue #2779, opened 2020-08-30
alicloud_db_instance call BuildStateConf parameter delay is too long?


zeromake:
The result is that no matter what the status is, I need to wait 5 minutes, No way to customize。

**BuildStateConf Method signature:**
```go
func BuildStateConf(pending, target []string, timeout, delay time.Duration, f resource.StateRefreshFunc)
```

https://github.com/aliyun/terraform-provider-alicloud/blob/b4393622069d2f0787a4e7c68c8809b8ff85802e/alicloud/resource_alicloud_db_instance.go#L269

And for expansion(instance_type, instance_storage), it will trigger two wait 10 minutes.
Actual expansion time is less than 2 minutes.

https://github.com/aliyun/terraform-provider-alicloud/blob/b4393622069d2f0787a4e7c68c8809b8ff85802e/alicloud/resource_alicloud_db_instance.go#L563

https://github.com/aliyun/terraform-provider-alicloud/blob/b4393622069d2f0787a4e7c68c8809b8ff85802e/alicloud/resource_alicloud_db_instance.go#L588

PR #2780, opened 2020-08-31
fix: db instance delay set is 30s

zeromake:
fix: #2779

### DIFF
--- a/alicloud/resource_alicloud_db_instance.go
+++ b/alicloud/resource_alicloud_db_instance.go
@@ -254,7 +254,7 @@ func resourceAlicloudDBInstanceCreate(d *schema.ResourceData, meta interface{}) 
 	d.SetId(response.DBInstanceId)
 
 	// wait instance status change from Creating to running
-	stateConf := BuildStateConf([]string{"Creating"}, []string{"Running"}, d.Timeout(schema.TimeoutCreate), 30*time.Second, rdsService.RdsDBInstanceStateRefreshFunc(d.Id(), []string{"Deleting"}))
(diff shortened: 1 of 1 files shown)


alicloud/resource_alicloud_db_instance.go
```diff
@@ -254,7 +254,7 @@ func resourceAlicloudDBInstanceCreate(d *schema.ResourceData, meta interface{})
 	d.SetId(response.DBInstanceId)
 
 	// wait instance status change from Creating to running
-	stateConf := BuildStateConf([]string{"Creating"}, []string{"Running"}, d.Timeout(schema.TimeoutCreate), 5*time.Minute, rdsService.RdsDBInstanceStateRefreshFunc(d.Id(), []string{"Deleting"}))
+	stateConf := BuildStateConf([]string{"Creating"}, []string{"Running"}, d.Timeout(schema.TimeoutCreate), 30*time.Second, rdsService.RdsDBInstanceStateRefreshFunc(d.Id(), []string{"Deleting"}))
```
xiaozhu36, line 257: I think the 30 seconds is too short. A new rds instance will cost at least 3 minutes. I think it can be change to 3*time.Minute.

xiaozhu36 (changes requested):

alicloud/resource_alicloud_db_instance.go
zeromake: @xiaozhu36 
This is the polling interval. Shouldn’t it be necessary to wait manually for 3 minutes before polling?

xiaozhu36: Yes, creating one instance will cost at least 3 min and there is no need to poll in the known pending stage. Too many polling will cost the API and user's throttling threshold.

xiaozhu36 (changes requested):

alicloud/resource_alicloud_db_instance.go
zeromake: @xiaozhu36 
already edited

PR #2785, opened 2020-09-02
Fix the managed_kubernetes docs.

zql-tqy:


website/docs/r/cs_managed_kubernetes.html.markdown
```diff
@@ -114,6 +114,11 @@ The following arguments are supported:
 
 #### Addons 
 It is a new field since 1.75.0. You can specific network plugin,log component,ingress component and so on.     
+
+-> **NOTE:** If you upgrade from a version before 1.90.0 to 1.90.0+(contains 1.90.0), you need to pay attention to the following notes:
+-> **NOTE:** If disabled is specified as `"false"` or `"true"`, you can directly upgrade；
+-> **NOTE:** If you specify disabled as `""`, you need to set disabled to `false` in the current version and execute `terraform apply` before upgrading;
+
```
xiaozhu36, line 122: -> **NOTE:** If you want to upgrade provider to 1.90.0+, you need to pay attention to the `disabled` value. If the value is `""`, you need to modify it to `"false"`, and then run `terraform apply` to make it effect. After that, you can modify the provider version to upgrade smoothly.

website/docs/r/cs_managed_kubernetes.html.markdown
```diff
@@ -238,7 +243,7 @@ It is a new field since 1.75.0. You can specific network plugin,log component,in
             {
                 "name"     = "nginx-ingress-controller",
                 "config"   = "",
-                "disabled": true,
+                "disabled" = "true",
```
xiaozhu36, line 246: "disabled" = "false",

xiaozhu36 (changes requested):

Issue #2845, opened 2020-09-19
Multi-AZ RDS database instance keeps being replaced


yannrouillard:
### Versions
Terraform v0.13.3
terraform-provider-alicloud 1.97.0

### Affected Resource(s)
- alicloud_db_instance

### Terraform Configuration Files
```hcl

resource "alicloud_db_instance" "test_database" {
  engine         = "MySQL"
  engine_version = "5.7"
  instance_name  = "test-db"

  zone_id      = "ap-southeast-5MAZ1(a,b)"
  vswitch_id   = "<ID_VSWITCH_A>,<ID_VSWITCH_B>"
  security_ips = []

  instance_type    = "rds.mysql.t1.small"
}
```

### Expected Behavior

When I create a highly-available RDS Instance in a multi-AZ zone, I expect subsequent launch of terraform not to detect any change nor to try to replace the instance.

### Actual Behavior

The second time I run `terraform`, `terraform` sees changes in the `vswitch_id` configuration:

```
 ~ vswitch_id                 = "<ID_VSWITCH_A>" -> "<ID_VSWITCH_A>,<ID_VSWITCH_B>" # forces replacement
```

and so tries to delete and recreate the database.

### Steps to Reproduce

Just use the terraform configuration I put above and run `terraform apply` twice.

yannrouillard:
What happens is that the RDS API `DescribeDBInstanceAttributes` of Alicloud always returns one vswitch in the `VSwitchId` field, the first vswitch given, even if 2 vswitches were given when the database has been created.

In fact the information about the second vswitch used is present nowhere in the output of `DescribeDBInstanceAttributes`.

So when comparing, `terraform` always detects a difference between the deployed and the expected configuration.

yannrouillard:
A possible workaround currently is to add:

```hcl
  lifecycle {
    prevent_destroy = true
    # Unfortunately this module is buggy and wants 2 vswitches at the first deployment
    # for a multi-az DB, and 1 vswitch for subsequent updates, otherwise it replaces the DB
    # we workaround that by ignoring any vswitch_id change...
    ignore_changes = [vswitch_id]
  }
```

chushenmeshile:
@yannrouillard Because the property is forcenew, once you modify it, they will delete the new one

yannrouillard:
No @chushenmeshile  it is not linked, I don't modify the vswitch option at all in my configuration.

chushenmeshile:
[...]

PR #2846, opened 2020-09-20
Fix/multi az database constant replacement

yannrouillard:
Here is a proposed fix for issue #2845.

There are 2 commits in this PR:

 - the first one just ignore any additional vswitches when computing the difference between actual state and expected state.
    this easily solves the issue, the drawback is that if someone changes the additional vswitch in the terraform configuration, it will never be detected.

 - the second commit solves this problem by adding a new field to store the additional vswitch `vswitch_id_slave_1`, this allows `terraform` to keep the information about the additional vswitch and detect any subsequent change. It seems to be better semantic anyway and it's aligned with the `ZoneIdSlave1` argument supported by the API (not yet by the terraform module).

Happy to discuss if this solution doesn't fit you.

### DIFF
--- a/alicloud/resource_alicloud_db_instance.go
+++ b/alicloud/resource_alicloud_db_instance.go
@@ -107,6 +107,11 @@ func resourceAlicloudDBInstance() *schema.Resource {
 					return old == Trim(strings.Split(new, COMMA_SEPARATED)[0])
 				},
 			},
+			"vswitch_id_slave_1": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+			},
 			"instance_name": {
 				Type:         schema.TypeString,
 				Optional:     true,
@@ -791,6 +796,10 @@ func buildDBCreateRequest(d *schema.ResourceData, meta interface{}) (*rds.Create
 	}
 
 	vswitchId := Trim(d.Get("vswitch_id").(string))
+
+	if vswitchIdSlave1, ok := d.GetOk("vswitch_id_slave_1"); ok && Trim(vswitchIdSlave1.(string)) != "" {
+		vswitchId = strings.Join([]string{vswitchId, vswitchIdSlave1.(string)}, ",")
+	}
 
 	request.InstanceNetworkType = string(Classic)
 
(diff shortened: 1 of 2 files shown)


alicloud/resource_alicloud_db_instance.go
```diff
@@ -101,6 +101,19 @@ func resourceAlicloudDBInstance() *schema.Resource {
...
+					return old == Trim(strings.Split(new, COMMA_SEPARATED)[0])
+				},
+			},
+			"vswitch_id_slaves": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				ForceNew: true,
+				Optional: true,
```
xiaozhu36, line 116: I think there is no need to add an extra field `vswitch_id_slaves` to support multi vswitch and just input several vswitches into `vswitch_id` with comma.

xiaozhu36 (changes requested):

alicloud/resource_alicloud_db_instance.go
yannrouillard: Hi @xiaozhu36,

Thanks for reviewing.

The problem with the multiple vswitches separated with comma is that we run into #2845 
and if we just compute diff on the first vswitch (like the first commit of this PR does), then any change to the additional vswitches are ignored.

yannrouillard: Hi @xiaozhu36 

Will you have some time soon to review this PR ?

yannrouillard: Hi @xiaozhu36

I am opened to make any required changes, but still waiting for your feedbacks.

xiaozhu36:
HI @yannrouillard Thanks for your contribution. Considering internal spec, we have supported this feature by PR #2924 and it has been published in the version 1.101.0. Please check it.

yannrouillard:
Hi @xiaozhu36 

Unfortunately I tried with version 1.101.0 and it still doesn't fix the original issue raised here.
My databases keep being replaced.

Anyway I will comment on the issue #2845 rather than here.

PR #2891, opened 2020-10-10
bugfix: implement DiffSuppressFunc for chart list of log dashboard and pass required fields on update

cofyc:
**What it does**

- implement DiffSuppressFunc to compare chart list semantically
  - JSON fields are unordered after marshaling, we cann't compare the raw string
- dashboard_name and char_list fields must be passed to the UPDATE API every time even if they are not changed
  - if we omit dashboard_name if it's not changed, it will be updated to dashboard id
  - chart_list is required field (the update request will fail)

### DIFF
--- a/alicloud/resource_alicloud_log_dashboard.go
+++ b/alicloud/resource_alicloud_log_dashboard.go
@@ -3,6 +3,9 @@ package alicloud
 import (
 	"encoding/json"
 	"fmt"
+	"log"
+	"reflect"
+	"sort"
 	"time"
 
 	sls "github.com/aliyun/aliyun-log-go-sdk"
@@ -38,8 +41,9 @@ func resourceAlicloudLogDashboard() *schema.Resource {
 				Optional: true,
 			},
 			"char_list": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:             schema.TypeString,
+				DiffSuppressFunc: jsonPolicyDiffSuppress,
+				Required:         true,
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


cofyc:
@xiaozhu36 PTAL

alicloud/resource_alicloud_log_dashboard.go
```diff
@@ -38,8 +41,9 @@ func resourceAlicloudLogDashboard() *schema.Resource {
 				Optional: true,
 			},
 			"char_list": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:             schema.TypeString,
+				DiffSuppressFunc: jsonPolicyDiffSuppress,
```
xiaozhu36, line 45: There is a stable diffsuppressfunc to support it and please refer to: https://github.com/aliyun/terraform-provider-alicloud/blob/master/alicloud/resource_alicloud_ram_policy.go#L67

xiaozhu36 (changes requested):

alicloud/resource_alicloud_log_dashboard.go
cofyc: `compareJsonTemplateAreEquivalent` can't work here. In my testing, `terrarform apply` show changes every time. 

Before comparing the chart lists, we must sort the charts.

cofyc: I guess the alicloud API reordered the chart list, so after the configurations are applied to the alicloud,  the`chart_list` may diff from the user configuration in the next `terraform apply` (using compareJsonTemplateAreEquivalent). 

However, the order of the chart list does not matter. I think it's better to sort the charts before comparing in this case.

PR #2986, opened 2020-11-04
fix network acl error

zhengjc2018:


alicloud/resource_alicloud_network_acl_entries_test.go
```diff
@@ -342,18 +341,18 @@ resource "alicloud_network_acl_entries" "default" {
...
       port = "-1/-1"
       destination_cidr_ip = "0.0.0.0/32"
       name = "${var.name}"
       entry_type = "custom"
       policy = "accept"
       description = "${var.name}"
-    },
-    {
+    }
+
```
elsnepal, line 354: what is the reason for requiring extra whitespace between the two egresses?

alicloud/resource_alicloud_network_acl_entries_test.go
```diff
@@ -323,17 +322,17 @@ resource "alicloud_network_acl" "default" {
...
       port = "-1/-1"
       source_cidr_ip = "0.0.0.0/32"
       name = "${var.name}"
       entry_type = "custom"
       policy = "accept"
       description = "${var.name}"
-    },
-    {
+    }
+
```
elsnepal, line 334: what is the reason for requiring extra whitespace between the two ingresses?

elsnepal (commented):
does this handle extra whitespace between the ingress/egress (as seen in your test case below; I have commented inline) when someone creates resources with dynamic for_each?

xiaozhu36 (approved):

Issue #3075, opened 2020-11-26
alicloud_db_account can't create super account


yongzhang:
Hi, 

Terraform version: 0.13.5
Provider version: 1.104.0

Steps to reproduce:
1. Create super account by `alicloud_db_account` with parameter `type = "Super"` and apply, succeed.
2. Re-run `terraform apply`, it will show below changes:

```
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # alicloud_db_account.mysql_super_user["dev-shared"] must be replaced
-/+ resource "alicloud_db_account" "mysql_super_user" {
      ~ id          = "rm-uf6aq4j8v55ooj87z:devops" -> (known after apply)
        instance_id = "rm-uf6aq4j8v55ooj87z"
        name        = "devops"
        password    = (sensitive value)
      ~ type        = "Normal" -> "Super" # forces replacement
    }

  # alicloud_db_account.postgresql_super_user["dev-shared"] must be replaced
-/+ resource "alicloud_db_account" "postgresql_super_user" {
      ~ id          = "pgm-uf67h8b5hwl9d883:devops" -> (known after apply)
        instance_id = "pgm-uf67h8b5hwl9d883"
        name        = "devops"
        password    = (sensitive value)
      ~ type        = "Normal" -> "Super" # forces replacement
    }

Plan: 2 to add, 0 to change, 2 to destroy.

...

alicloud_db_account.mysql_super_user["dev-shared"]: Destroying... [id=rm-uf6aq4j8v55ooj87z:devops]
alicloud_db_account.postgresql_super_user["dev-shared"]: Destroying... [id=pgm-uf67h8b5hwl9d883:devops]
alicloud_db_account.postgresql_super_user["dev-shared"]: Destruction complete after 1s
alicloud_db_account.postgresql_super_user["dev-shared"]: Creating...
alicloud_db_account.mysql_super_user["dev-shared"]: Destruction complete after 2s
alicloud_db_account.mysql_super_user["dev-shared"]: Creating...
alicloud_db_account.postgresql_super_user["dev-shared"]: Creation complete after 2s [id=pgm-uf67h8b5hwl9d883:devops]
alicloud_db_account.mysql_super_user["dev-shared"]: Creation complete after 2s [id=rm-uf6aq4j8v55ooj87z:devops]

Apply complete! Resources: 2 added, 0 changed, 2 destroyed.
```

From console, it actually created "Normal" account not "Super":
[...]

PR #3076, opened 2020-11-26
fix var name for account type in alicloud_db_account 

yongzhang:
This is to fix https://github.com/aliyun/terraform-provider-alicloud/issues/3075


yongzhang:
@chushenmeshile @xiaozhu36 please review, thanks.

PR #3341, opened 2021-03-15
Fix DescribeKeyPairAttachment #3338

PMExtra:
Fix #3338

Issue #3338, opened 2021-03-12
ECS - key_pair_attachment bug if key_pair name contains ':'.


PMExtra:
### Terraform Version
v0.14.8

### Affected Resource(s)
- alicloud_key_pair_attachment

### Terraform Configuration Files
```hcl
resource "alicloud_key_pair" "key" {
  key_name   = "示例:测试"
  public_key = "..."
}
resource "alicloud_key_pair_attachment" "test" {
  key_name     = alicloud_key_pair.test_key.key_name
  instance_ids = [alicloud_instance.test_instance.id]
}
```

### Output
```plain
Error: [ERROR] terraform-provider-alicloud/alicloud/resource_alicloud_key_pair_attachment.go:113:
[ERROR] terraform-provider-alicloud/alicloud/common.go:803:
Invalid Resource Id 示例:测试:["i-xxxxxxxxxx"]. Expected parts' length 2, got 3
```

### Reference
https://github.com/aliyun/terraform-provider-alicloud/blob/52a6b6e31060a18666e89794eae7651e3508bdbc/alicloud/resource_alicloud_key_pair_attachment.go#L106-L114
https://github.com/aliyun/terraform-provider-alicloud/blob/91e52950a511dfd70d611d611fa429b91cba441a/alicloud/service_alicloud_ecs.go#L505-L507
https://github.com/aliyun/terraform-provider-alicloud/blob/13b5caa804f0096d5006549178f90171d3774da4/alicloud/common.go#L799-L806

It expect 2 colon here, but the key_pair ID may contains colon. In this case, the count of colon will be more than 2.

It should check at least one colon, and split by the last colon.

PR #3395, opened 2021-03-31
resource/alicloud_keypair_attachment: Fix KeyPairAttachment bug when the key pair name contains colon

PMExtra:
About #3338 

The last pull request (#3341) did NOT fix this issue completely. It only handle the colon when describe the ECS, but miss the case when read or delete the KeyPairAttachment.


PMExtra:
@xiaozhu36 Please take a look.

PR #3412, opened 2021-04-06
update: remove default value of spec

zhengjc2018:


xiaozhu36:
It has been fixed by #3458

PR #3427, opened 2021-04-12
resource/alicloud_vswitch: Adds ipv6_cidr_block

ezio-ballarin:


xiaozhu36:
HI @ezio-ballarin Can you add the testcase for this feature?

ezio-ballarin:
hey @xiaozhu36, are these tests OK? I couldn't find where to enable IPv6 for the VPC in the `resource_alicloud_vswitch_test.go` file. I need the default VPC to be IPv6 enabled, could you help me verify or find where to edit?

ezio-ballarin:
hi again @xiaozhu36 it's been a couple months, any update here?

xiaozhu36:
HI @ezio-ballarin I am sorry for late reply. Did you run this your testcase? I think there is any OpenAPI bug and this bug can be found by running the testcase.
The bug is `ipv6_cidr_block`type  is `int` in the [CreateVSwitch](https://help.aliyun.com/document_detail/35745.html) but in the [DescribeVSwitchAttributes](https://help.aliyun.com/document_detail/94567.html) is `string`. It will fail when running `d.Set(ipv6_cidr_block)` and have diff when running the next terraform command. 
It is the OpenAPI. bug and I have submitted this issue to product team to fix it. You can also create a ticket to push them.

ezio-ballarin:
No worries @xiaozhu36, we are all very busy so I understand. I've read through all the docs in this repo and I am probably not understanding enough, but is "acceptance testing" what I need to do for this? It sounds like I will need to run the acceptance tests.

Also, I will look into the OpenAPI type issue you pointed out and talk with my TAMs about it.

LiangZugeng:
@xiaozhu36  I don't think it's a bug, it's designed as is.

1. When creating the switch, the first 56-bit of IPv6 address was inherited from the VPC, therefore it only requires a number (0-255) as the last 8-bit of the IPv6 subnet (::/64).

2. The IPv6CidrBlock attribute in DescribeVSwitchAttribute is a string which combines the first 56-bit from the VPC and the 8-bit of the VSwitch to show the full IPv6 subnet.

I'd like to propose a solution: Add a new int type argument called ipv6_cidr_block_suffix and make ipv6_cidr_block an string attribute. This is aligned with the current design of the Aliyun API.

xiaozhu36:
@LiangZugeng Thanks your ideas. I think the `ipv6_cidr_block_suffix` is not good for the issue. The vswitch ipv6 count is equals to the VPC ipv6_cidr_block suffix subtracts the VSwitch ipv6_cidr_block suffix. So, I think there should describe the VPC ipv6_cidr_block and do that when[ the resource setting the vswitch ipv6_cidr_block](https://github.com/aliyun/terraform-provider-alicloud/pull/3427/files#diff-d9d44fcb2ca064564e4b57d5dfee1e2b46fb4f2f7cb2a334e914ad38a845f0f1R162).

xiaozhu36:
Also, please @ezio-ballarin adds the attribute ipv6_cidr_block description in the resource docs.

LiangZugeng:
@xiaozhu36 Thanks for your prompt response, could you please elaborate it? I'm not sure I fully understand "I think there should describe the VPC ipv6_cidr_block and do that when the resource setting the vswitch ipv6_cidr_block."

Did you mean that there should be only a single argument called ipv6_cidr_block? If that's the case, what's the type of the argument? It will be great if a more detailed design can be suggested.

PR #3576, opened 2021-05-25
resource/alicloud_log_store: Adds new attribute encrypt_conf to support encrypt logstore

HuiJing-C:


alicloud/resource_alicloud_log_store.go
```diff
@@ -96,6 +96,49 @@ func resourceAlicloudLogStore() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"encrypt_conf": {
+				Type:     schema.TypeSet,
+				ForceNew: true,
+				Optional: true,
+				MaxItems: 1,
+				Default:  nil,
```
xiaozhu36, line 104: If there is no default, please remove it, not set it to nil

alicloud/resource_alicloud_log_store_test.go
```diff
@@ -123,6 +123,60 @@ func TestAccAlicloudLogStore_basic(t *testing.T) {
...
+				}),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheck(map[string]string{
+						"name":           name,
+						"project":        name,
+						"shard_count":    "1",
+						"encrypt_conf.#": "1",
+					}),
+				),
+			},
```
xiaozhu36, line 175: Please add ImportTestcase

xiaozhu36 (changes requested):

PR #3641, opened 2021-06-18
 NewResource `alicloud_event_bridge_schema_group`  NewDataSource `alicloud_event_bridge_schema_groups`

liuri1205:
NewDataSource `alicloud_event_bridge_schema_groups`

### DIFF
--- a/alicloud/provider.go
+++ b/alicloud/provider.go
@@ -436,6 +436,7 @@ func Provider() terraform.ResourceProvider {
 			"alicloud_log_stores":                                  dataSourceAlicloudLogStores(),
 			"alicloud_event_bridge_service":                        dataSourceAlicloudEventBridgeService(),
 			"alicloud_event_bridge_event_buses":                    dataSourceAlicloudEventBridgeEventBuses(),
+			"alicloud_event_bridge_schema_groups":                  dataSourceAlicloudEventBridgeSchemaGroups(),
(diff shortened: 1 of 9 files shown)


alicloud/provider.go
```diff
@@ -1161,6 +1166,13 @@ func endpointsSchema() *schema.Schema {
 		Optional: true,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
+				"eventbridge_share": {
```
xiaozhu36, line 1169: I think the endpoint name should be `event_bridge`

alicloud/service_alicloud_eventbridge_share.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"time"
+
+	"github.com/PaesslerAG/jsonpath"
+	util "github.com/alibabacloud-go/tea-utils/service"
+	"github.com/aliyun/terraform-provider-alicloud/alicloud/connectivity"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+type Eventbridge_shareService struct {
```
xiaozhu36, line 13: I think the service name should be EventBridgeService

xiaozhu36 (changes requested):

PR #3912, opened 2021-08-26
fix: `cs_managed_kubernetes` - kubeconfig file create folders and use correct fs.FileMode

hayorov:
if `cs_managed_kubernetes .kubeconfig` has value /foo/bar/.secret, ioutil.WriteFile failes with "No such file or directory"

Fixes:
- create intermediate folders (if required)
- Use correct fs.FileMode value

### DIFF
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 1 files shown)


hayorov:
@xiaozhu36

hayorov:
Hello, up!

PR #4535, opened 2022-01-26
testcase: Adds new unit test case for resource `alicloud_nas_lifecycle_policy`

super-eggs:
testcase: Adds new unit test case for resource
`alicloud_nas_lifecycle_policy`

### DIFF
--- a/alicloud/resource_alicloud_nas_lifecycle_policy_test.go
+++ b/alicloud/resource_alicloud_nas_lifecycle_policy_test.go
@@ -2,6 +2,16 @@ package alicloud
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


alicloud/resource_alicloud_nas_lifecycle_policy_test.go
```diff
@@ -86,3 +96,312 @@ resource "alicloud_nas_file_system" "default" {
 }
 `, name)
 }
+
+func TestAccAlicloudNASLifecyclePolicy_unit(t *testing.T) {
```
super-eggs, line 100: 单测96.8%覆盖率

PR #4805, opened 2022-03-29
testcase: Adds new unit test case for resource alicloud_dfs_file_system alicloud_dfs_access_group alicloud_dfs_access_rule

thisnihonglei:


alicloud/resource_alicloud_dfs_file_system_test.go
```diff
@@ -211,3 +219,259 @@ locals {
 }
 `, name)
 }
+
+func TestAccAlicloudDFSFileSystem_unit(t *testing.T) {
```
thisnihonglei, line 223: 单测覆盖率96.7%

PR #5045, opened 2022-06-08
resource/alicloud_cs_kubernetes_node_pool: fix-bug: nodepool labels and taints cannot be annotated as empty.

broune001:


broune001 (commented):
local test passed

PR #5084, opened 2022-06-20
New Datasource `alicloud_cdn_blocked_regions`;

super-eggs:
New Datasource `alicloud_cdn_blocked_regions`;


alicloud/data_source_alicloud_cdn_blocked_regions.go
```diff
@@ -0,0 +1,107 @@
+package alicloud
+
+import (
```
xinmu0426, line 3: 冲突了

PR #5092, opened 2022-06-22
resource/alicloud_fc_trigger,data_source/alicloud_fc_trigger: support eventbridge trigger type

git-qfzhang:
… `eventbridge` trigger type


alicloud/resource_alicloud_fc_trigger.go
```diff
@@ -83,6 +92,19 @@ func resourceAlicloudFCTrigger() *schema.Resource {
 
 						return resolvedOld == resolvedNew
 					}
+
+					if d.Get("type").(string) == string(fc.TRIGGER_TYPE_EVENTBRIDGE) {
```
xinmu0426, line 96: filed check ci没过

PR #5330, opened 2022-08-23
testcase: fix issue for acc test of alb system security policies

minchaow:
This PR is to fix the following issue for acc test of alb system security policies.
`
testing.go:635: Step 0 error: config is invalid: Invalid reference: A reference to a resource type must be followed by at least one attribute access, specifying the resource name.
`


minchaow:
@xiaozhu36 please kindly have a review if your time is available, thanks

PR #5406, opened 2022-09-12
resource/alicloud_cs_managed_kubernetes: Fix attributes bug caused by `rrsa_metadata`

Pangjiping:


Pangjiping:
related to #5375

PR #5611, opened 2022-11-23
resource/alicloud_cs_kubernetes_addon: Optimize components uninstall logic

saberrey:
1. Optimize components uninstall logic.
2. Adjust the scaling logic of the node pool to prevent the inconsistency between the nodecount and the desiredsize after the pop-up node fails for some reasons.
3. Optimize documents.


alicloud/resource_alicloud_cs_kubernetes.go
```diff
@@ -266,6 +266,7 @@ func resourceAlicloudCSKubernetes() *schema.Resource {
 			"worker_instance_charge_type": {
 				Type:         schema.TypeString,
 				Optional:     true,
+				Default:      common.PostPaid,
```
xiaozhu36, line 269: 如果有默认值，直接设置为 Computed

xiaozhu36 (changes requested):

PR #5984, opened 2023-03-14
resource/alicloud_cms_site_monitor: Fix English lower case bug.

thisnihonglei:


alicloud/resource_alicloud_cms_site_monitor_test.go
```diff
@@ -140,6 +140,36 @@ func TestAccAlicloudCmsSiteMonitor_update(t *testing.T) {
 	})
 }
 
+func TestAccAlicloudCmsSiteMonitor_basic1(t *testing.T) {
```
thisnihonglei, line 143: image

PR #6125, opened 2023-05-08
resource/alicloud_nlb_listener: Removes the limition for attribute security_policy_id and supports setting custom security policies

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6139, opened 2023-05-12
resource/alicloud_db_instance: Adds new attributes status and create_time; Fixes the connection string duplicate error caused by setting port to 3306 when creating

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6197, opened 2023-05-31
resource/alicloud_log_store: Enlarges the max_split_shard_count max valid value to 256

xiaozhu36:
…


xiaozhu36 (commented):
approved

xiaozhu36 (commented):
approved

PR #6242, opened 2023-06-18
resource/alicloud_vpc_peer_connection: Fixes the ResourceNotFound.InstanceId error when destroying it

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6245, opened 2023-06-19
resource/alicloud_log_project: Ignores the system tag which starting with acs for the attribute tags

xiaozhu36:
…


xiaozhu36 (commented):
approved

xiaozhu36 (commented):
approved

PR #6260, opened 2023-06-26
resource/alicloud_ga_basic_accelerator: Added the field tags

MrWolong:


alicloud/resource_alicloud_ga_basic_accelerator_test.go
```diff
@@ -53,12 +53,19 @@ func TestAccAlicloudGaBasicAccelerator_basic0(t *testing.T) {
 					"auto_use_coupon":        "true",
 					"auto_renew":             "false",
 					"auto_renew_duration":    "1",
+					"tags": map[string]string{
```
xiaozhu36, line 56: 一样

xiaozhu36 (changes requested):

alicloud/resource_alicloud_ga_basic_accelerator_test.go
MrWolong: 已修复

xiaozhu36 (approved):

PR #6290, opened 2023-07-03
resource/alicloud_ess_scaling_configuration: fixes the ServiceUnavailable error when updating the data_disk

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6310, opened 2023-07-13
resource/alicloud_bastionhost_instance: Fixes the ResourceNotFound error when using international account

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6451, opened 2023-08-28
resource/alicloud_cloud_firewall_instance: Adds new attributes cfw_account and account_number, and removes the attribute cfw_service

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6468, opened 2023-09-03
testcase: Adds or improves sweeper testcase for ssl_certificate_service_certificate and the resource group

xiaozhu36:


xiaozhu36 (commented):
approved

PR #6485, opened 2023-09-07
docs: Mark deprecated resource waf_instance,waf_domain

susu9527:


website/docs/r/waf_domain.html.markdown
```diff
@@ -9,6 +9,9 @@ description: |-
 
 # alicloud\_waf\_domain
 
+-> **DEPRECATED:**  This resource is based on Config's old version OpenAPI, and it has been deprecated from version `1.211.0`.
```
xiaozhu36, line 12: 直接写：This resource has been deprecated and using xxx instead.

xiaozhu36 (changes requested):

website/docs/r/waf_domain.html.markdown
susu9527: 已改

PR #6498, opened 2023-09-13
resource/alicloud_instance: Adds new output attribute system_disk_id, and fixes the resource not found error when there is missing system disk; resource/alicloud_ecs_instance_set: Fixes the resource not found error when there is missing system disk

xiaozhu36:
…


xiaozhu36 (commented):
approved

xiaozhu36 (commented):
approved

xiaozhu36 (commented):
approved

xiaozhu36 (commented):
approved

PR #6508, opened 2023-09-15
resource/alicloud_ess_scalinggroup: Fixes the BackendServer.configuring error when attaching vserver groups

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6511, opened 2023-09-16
resource/alicloud_slb_listener: Fixes the VServerGroupId does not exist error when updating listener attribute

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6518, opened 2023-09-19
resource/alicloud_kvstore_instance: Ignores the period diff when payment type is PostPaid

xiaozhu36:
…


xiaozhu36 (commented):
approved

xiaozhu36 (commented):
approved

PR #6545, opened 2023-09-27
resource/alicloud_gpdb_instance: Fixes the import diff error caused by attribute security_ip_list

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6551, opened 2023-09-28
resource/alicloud_mongodb_instance: Fixes the StorageTypeOrInstanceTypeNotSupported error when refreshing the resource state

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6552, opened 2023-09-28
resource/alicloud_ecs_key_pair: Fixes the InvalidResourceGroup.NotFound error when resource_group_id is empty

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #6792, opened 2023-12-24
resource/alicloud_dms_enterprise_instance: mark use_dsql as computed property; testcase: fix dms testcase.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #6800, opened 2023-12-26
testcase: fix mongodb testcae; resource/alicloud_mongodb_sharding_instance: mark vswitch_id as Computed property.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #6841, opened 2024-01-05
resource/alicloud_alb_server_group: add retry error code; resource/alicloud_alb_rule: validation for protocol and vpc_id while server_group_type is Fc.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #6860, opened 2024-01-15
resource/alicloud_instance: add new type for system_disk_category;datasource/alicloud_instance_types: add new type for system_disk_category.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #6871, opened 2024-01-16
resource/alicloud_alb_server_group: Supported protocol set to gRPC, and health_check_config.health_check_method set to POST

MrWolong:


ChenHanZhang (commented):
approved

PR #6975, opened 2024-02-18
resource/alicloud_hbr_oss_backup_plan: fix bug while leave this property empty for field prefix.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #6984, opened 2024-02-19
resource/alicloud_mse_engine_namespace: add parameter for create, update and delete api.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #6990, opened 2024-02-20
data-source/alicloud_alb_rules: Fixed the panic error caused by type of rules.0.rule_actions.0.redirect_config.0.port

MrWolong:


ChenHanZhang (commented):
approved

PR #7003, opened 2024-02-22
resource/alicloud_gpdb_instance: Added retry strategy for error code OperationDenied.OrderProcessing; Fixed the create error caused by state refresh

MrWolong:


ChenHanZhang (commented):
approved

PR #7059, opened 2024-03-15
resource/alicloud_ecs_launch_template: remove default value for system_disk.performance_level.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7060, opened 2024-03-15
resource/alicloud_quotas_*: Support for international languages; data-source/alicloud_quotas_*: Support for international languages.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7201, opened 2024-04-28
resource/alicloud_log_store: optimize Metrics telemetry type code implementation while create and update; resource/alicloud_log_project: Add constraint for project_name.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7214, opened 2024-05-07
resource/alicloud_click_house_db_cluster: Added support for cluster version 23.8, added support for in-place cluster storage upgrade. resource/alicloud_click_house_account: Added support for creating super account.

itkillsnone:
resource/alicloud_click_house_db_cluster: Added support for cluster version 23.8, added support for in-place cluster storage upgrade.
resource/alicloud_click_house_account: Added support for creating super account.


ChenHanZhang (commented):
approved

PR #7252, opened 2024-05-16
resource/alicloud_vpc: add retry for DependencyViolation.VSwitch; data-source/alicloud_route_tables: add retry for throttling

shanye997:


ChenHanZhang (commented):
approved

PR #7320, opened 2024-06-03
resource/alicloud_ots_table: support new sse type ByOk and allow_update param; resource/alicloud_ots_search_index: fix optional bug of index_setting and index_sort

waterxjw:
resource/alicloud_ots_table: support new sse type ByOk


ChenHanZhang (commented):
approved

PR #7322, opened 2024-06-03
resource/alicloud_image_import: Added the field boot_mode; Improved alicloud_image_import testcase

MrWolong:


ChenHanZhang (commented):
approved

PR #7334, opened 2024-06-06
resource/alicloud_mongodb_sharding_instance: Added the field storage_type; Removed the ForceNew for field engine_version; Supported for new action UpgradeDBInstanceEngineVersion

MrWolong:


ChenHanZhang (commented):
approved

PR #7360, opened 2024-06-19
resource/alicloud_click_house_db_cluster: Added support for in-place cluster node group and class upgrade.

itkillsnone:
resource/alicloud_click_house_db_cluster: Added support for in-place cluster node group and class upgrade.


ChenHanZhang (commented):
approved

ChenHanZhang (commented):
approved

PR #7404, opened 2024-07-01
resource/alicloud_ecs_launch_template: add new attribute auto_renew, auto_renew_period and period_unit.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7428, opened 2024-07-05
docs: Imporved targets parameter description for cms_alarm.

simonrainerbohmer:
Added an update to clarify the description for the targets parameter.


ChenHanZhang:
Thanks for the contribution

PR #7469, opened 2024-07-19
resource/alicloud_ess_eci_scaling_configuration: Fix cpu_options_core and cpu_options_threads_per_core is zero.

fuliu-zln:
… and cpu_options_threads_per_core is zero.


ChenHanZhang (commented):
approved

PR #7559, opened 2024-08-13
resource/alicloud_mongodb_instance: Added the field provisioned_iops; Supported storage_type set to cloud_auto; Removed the ForceNew for field storage_type; Supported for new action ModifyDBInstanceDiskType

MrWolong:


MrWolong (commented):
approved

ChenHanZhang (commented):
approved

PR #7565, opened 2024-08-15
resource/alicloud_instance: Fixed the create bug when only creating a primary networkInterface

MrWolong:


MrWolong (commented):
approved

MrWolong (commented):
approved

ChenHanZhang (commented):
approved

PR #7617, opened 2024-09-02
resource/alicloud_gpdb_account: add new attribute account_type, database_name; resource/alicloud_gpdb_db_resource_group: add new attribute role_list; resource/alicloud_gpdb_remote_adb_data_source: support update attribute user_name and user_password.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7622, opened 2024-09-03
resource/alicloud_hbr_policy_binding: add new attribute cross_account_role_name, cross_account_type and cross_account_user_id.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7684, opened 2024-09-19
resource/alicloud_route_entry: Added the field description; Improved alicloud_route_entry testcase

MrWolong:


MrWolong (commented):
approved

PR #7730, opened 2024-10-15
resource/alicloud_ga_endpoint_group: Added the field sub_address; Supported type set to ALB, NLB, ENI, OSS

MrWolong:


MrWolong (commented):
approved

PR #7770, opened 2024-10-23
resource/alicloud_eais_instance: Fixed the import error caused by field vswitch_id and security_group_id no value; Fixed the diff error caused by field instance_name

MrWolong:


MrWolong (commented):
approved

MrWolong (commented):
approved

PR #7781, opened 2024-10-28
resource/alicloud_vswitch: Added the field is_default; Changed the field vpc_id, cidr_block from Required to Optional; Supported for new action CreateDefaultVSwitch

MrWolong:


MrWolong (commented):
approved

PR #7808, opened 2024-11-05
resource/alicloud_cen_transit_router_vbr_attachment: Fixed the diff error caused by field resource_type; Improved alicloud_cen_transit_router_vbr_attachment testcase

MrWolong:


MrWolong (commented):
approved

PR #7845, opened 2024-11-13
resource/alicloud_cloud_sso_access_configuration: Fixed the diff error caused by field permission_policies.permission_policy_document

MrWolong:


MrWolong (commented):
approved

PR #7857, opened 2024-11-14
resource/alicloud_fcv3_trigger: add new attributes http_trigger, last_modified_time and target_arn.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #7910, opened 2024-11-27
 resource/alicloud_ess_scaling_rule: optimize query rule

fuliu-zln:
resource/alicloud_ess_scaling_rule: optimize query rule


ChenHanZhang (commented):
approved

PR #7936, opened 2024-12-04
resource/alicloud_sae_application: Fixed the diff error caused by field micro_registration

MrWolong:


MrWolong (commented):
approved

MrWolong (commented):
approved

PR #7953, opened 2024-12-06
resource/alicloud_data_works_project: add new attributes dev_environment_enabled, dev_role_disabled, pai_task_enabled and etc.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8053, opened 2024-12-27
resource/alicloud_oss_bucket_logging: add wait for status while create and update opertion.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8062, opened 2025-01-02
resource/alicloud_snat_entry: Refactored the resource alicloud_snat_entry; Added the field eip_affinity; Removed the ForceNew for field snat_ip

MrWolong:


MrWolong (commented):
approved

PR #8078, opened 2025-01-04
resource/alicloud_oss_bucket: Fixes the panic error when not setting bucket name explicitly

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #8104, opened 2025-01-08
resource/alicloud_message_service_topic: Refactored the resource alicloud_message_service_topic; Added the field enable_logging, tags, create_time; Deprecated the field logging_enabled

MrWolong:


MrWolong (commented):
approved

PR #8106, opened 2025-01-08
resource/alicloud_ecs_session_manager_status: Fixed the create, update error caused by state refresh

MrWolong:


MrWolong (commented):
approved

PR #8235, opened 2025-02-03
resource/alicloud_actiontrail_trail: Add missing params to export trail to maxcompute

frco9:
As documented in API doc https://next.api.alibabacloud.com/api/Actiontrail/2020-07-06/CreateTrail?RegionId=cn-beijing We can now create a trail that writes in a MaxCompute project. This PR adds these changes

### DIFF
--- a/alicloud/resource_alicloud_actiontrail_trail.go
+++ b/alicloud/resource_alicloud_actiontrail_trail.go
@@ -68,6 +68,14 @@ func resourceAlicloudActiontrailTrail() *schema.Resource {
 				Computed: true,
 				Optional: true,
 			},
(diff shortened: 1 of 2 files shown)


xiaozhu36:
Hi @frco9 Thanks for your attribution. There is missing new field description in the docs, and missing new accepting test case for the new field. Could you please add them?

frco9:
Hello ! 
Yes sure I can add them. However I've struggle to make tests pass locally. 
I would have expected Unit test to run without Alicloud credentials but they are failing if I'm not providing proper credentials. 

Concerning the acceptance test, I don't really have an AliCloud account where I could make them run. How is it working in that case ?

PR #8248, opened 2025-02-05
resource/alicloud_ecs_deployment_set: Deprecated the field domain, granularity; Fixed the diff error caused by field strategy, domain, granularity

MrWolong:


MrWolong (commented):
approved

PR #8309, opened 2025-02-10
resource/alicloud_alb_load_balancer: add new attribute zone_mappings.intranet_address, zone_mappings.allocation_id and etc.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8338, opened 2025-02-12
client: more product client supports refresh credentials automatically and improves the getting endpoint method

xiaozhu36:
…


xiaozhu36 (commented):
approved

Issue #8365, opened 2025-02-17
alicloud_mse_nacos_config importing not support colons in dataIds currently


luoxiner:
We can not import alicloud_mse_nacos_config when dataId contains ":", e.g.
`tf import alicloud_mse_nacos_config.example test:test:a::b:c`
a::b is dataId and c is group

PR #8366, opened 2025-02-17
resource/alicloud_mse_nacos_config: Support import nacos config when dataId contains colons; data-source/alicloud_mse_nacos_configs: Escape dataId, group, namespace when build Id

luoxiner:
to #8365


ChenHanZhang (commented):
approved

PR #8370, opened 2025-02-18
resource/alicloud_nlb_server_group: Add new attribute health_check_req and health_check_exp.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8453, opened 2025-03-04
resource/alicloud_ros_stack: Fixes the Throttling.User error when refreshing the resource state

xiaozhu36:
…


xiaozhu36 (commented):
approved

PR #8462, opened 2025-03-06
resource/alicloud_selectdb_db_instance: Add attributes of admin_pass, support creating 4.0 instance.

Kamosphere:
… support creating 4.0 instance.


MrWolong (commented):
approved

PR #8478, opened 2025-03-10
resource/alicloud_ebs_disk_replica_pair: Add new attributes one_shot, reverse_replicate and tags.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8524, opened 2025-03-19
resource/alicloud_havip_attachment: Fix bug while havip count larger 20; testcase: Improves testcase for havip_attachment.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8544, opened 2025-03-24
resource/alicloud_express_connect_virtual_border_router: Fixed the diff error caused by field vlan_id

MrWolong:


MrWolong (commented):
approved

PR #8549, opened 2025-03-25
resource/alicloud_resource_manager_shared_resource: Remove enumeration check for resource_type.

yaoyv001:


ChenHanZhang (commented):
approved

PR #8552, opened 2025-03-25
resource/alicloud_api_gateway_instance: Fixed error when ingress_vpc_id is not set

pengjinqiang:
resource/alicloud_api_gateway_instance: Fixed error when ingress_vpc_id is not set


ChenHanZhang (commented):
approved

PR #8559, opened 2025-03-26
resource/alicloud_rocketmq_consumer_group: Added the field max_receive_tps, region_id, consume_retry_policy.dead_letter_target_topic

MrWolong:


MrWolong (commented):
approved

PR #8570, opened 2025-03-27
resource/alicloud_api_gateway_group: Supports enable vpc intranet domain.

pengjinqiang:
resource/alicloud_api_gateway_group: Supports enable vpc intranet domain.


ChenHanZhang (commented):
approved

PR #8614, opened 2025-04-10
resource/alicloud_nlb_server_group_server_attachment: Fixed the issue where the update operation accidentally set the weight to 0 when the weight is not specified.

ChenHanZhang:


ChenHanZhang (commented):
approved

ChenHanZhang (commented):
approved

PR #8686, opened 2025-05-08
resource/alicloud_kms_instance: support modify instance name and fix bug while delete Subscription instance.

ChenHanZhang:


ChenHanZhang (commented):
approved

ChenHanZhang (commented):
approved

PR #8705, opened 2025-05-14
resource/alicloud_ecs_network_interface: Fixed the diff error caused by field security_group_ids, security_groups

MrWolong:


MrWolong (commented):
approved

PR #8756, opened 2025-06-06
resource/alicloud_dfs_file_system: Fix bug while create file_system with partition_number 0; testcase: Fix cases for dfs_vsc_mount_point.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8781, opened 2025-06-19
resource/alicloud_ram_role: Refactored the resource alicloud_ram_role; Added the field role_name, assume_role_policy_document, tags, create_time; Deprecated the field name, document

MrWolong:


MrWolong (commented):
approved

PR #8807, opened 2025-07-01
resource/alicloud_kms_secret: Supported secret_type set to Redis, PolarDB; Fixed the create error caused by setting secret_type to RAMCredentials

MrWolong:


MrWolong (commented):
approved

MrWolong (commented):
approved

PR #8836, opened 2025-07-17
resource/alicloud_vpc: fix bug while create vpc with system_route_table_route_propagation_enable valued true.

ChenHanZhang:


ChenHanZhang (commented):
approved

PR #8882, opened 2025-08-05
resource/alicloud_mse_cluster: add resource version upgrade capability

YangzeXie:
A version_code field has been added to identify the specific instance version. When version_code is LATEST, the system checks whether the instance is the latest version and plans an upgrade. If version_code is a specific version number, the system compares the version number to see if it has changed. If so, the system checks whether the instance is the latest version. If it meets the requirements, an upgrade is triggered.
MSE cluster resource restrictions only support upgrades to the latest version; upgrades to a specific version are not supported. The logic is the same as in the MSE console.


ChenHanZhang (commented):
approved

ChenHanZhang (commented):
approved

PR #8921, opened 2025-08-20
resource/alicloud_hologram_instance: Refactored the resource alicloud_hologram_instance; Added the field region_id; Supported instance_type set to Serverless

MrWolong:


MrWolong (commented):
approved

PR #8924, opened 2025-08-21
resource/alicloud_alb_server_group: Fixed the diff error caused by setting server_group_type to Fc

MrWolong:


MrWolong (commented):
approved

PR #8961, opened 2025-09-03
resource/alicloud_ddosbgp_ip: Refactored the resource alicloud_ddosbgp_ip; Deprecated the field resource_group_id

MrWolong:


MrWolong (commented):
approved

PR #9085, opened 2025-10-22
resource/alicloud_cms_site_monitor: Added the field option_json; Deprecated the field options_json

MrWolong:


MrWolong (commented):
approved

PR #9123, opened 2025-11-03
resource/alicloud_alikafka_topic: Refactored the resource alicloud_alikafka_topic; Added the field configs, status, region_id, create_time

MrWolong:


MrWolong (commented):
approved

PR #9126, opened 2025-11-05
resource/alicloud_alb_server_group: Fix bug while set servers more than 20 instances.

ChenHanZhang:


ChenHanZhang (commented):
approved

ChenHanZhang (commented):
approved

PR #9156, opened 2025-11-12
resource/alicloud_eflo_vpd: Refactored the resource alicloud_eflo_vpd; Added the field secondary_cidr_blocks, tags, region_id; Removed the ForceNew for field resource_group_id

MrWolong:


MrWolong (commented):
approved

PR #9399, opened 2026-01-15
resource/alicloud_rocketmq_instance: Fixed the creation error caused by field product_info.send_receive_ratio

MrWolong:


MrWolong (commented):
approved

PR #9586, opened 2026-03-19
resource/alicloud_resource_manager_shared_resource: Added the field properties_resource_arn, resource_property

MrWolong:


MrWolong (commented):
approved

PR #9634, opened 2026-04-14
resource/alicloud_drds_instance: set required attributes; resource/alicloud_lindorm_instance: update validation; resource/alicloud_ocean_base_instance: update validation

shanye997:


shanye997 (commented):
approved

PR #9271, opened 2025-12-04
resource/alicloud_cs_kubernetes_node_pool: Added the field auto_mode, instance_metadata_options etc; resource/alicloud_cs_managed_kubernetes: Mark deletion_protection as Computed.

ChenHanZhang:

PR #9711, opened 2026-05-05
Resource/docs: Allow AliyunLinux4 images as image_type on k8s nodepools

grammeaway:
## Current state

The `image_type` field on `alicloud_cs_kubernetes_node_pool` validates user input client-side via a `StringInSlice`. As of today, the v4 family is represented by exactly one string: `AliyunLinux4ContainerOptimized` (added in #9271  ). There is no plain `AliyunLinux4` (x86), no `AliyunLinux4Arm64`, and no `AliyunLinux4Security`, even though the corresponding v3 entries (`AliyunLinux3`, `AliyunLinux3Arm64`, `AliyunLinux3ContainerOptimized`) all exist.

The resource and data-source docs also drift from the schema:

- `website/docs/r/cs_kubernetes_node_pool.html.markdown` is missing bullets for two values the schema already accepts: `AliyunLinux Qboot` and `AliyunLinuxSecurity`.
- `website/docs/d/cs_kubernetes_node_pools.html.markdown` is missing five (the two above plus `AliyunLinux3ContainerOptimized`, `Custom`, `AliyunLinux4ContainerOptimized`).

## Why this matters
I needed to run an ACK node pool on Alibaba Cloud Linux 4 (ARM64, `ecs.g8y.large`). Setting `image_type = "AliyunLinux4Arm64"` failed with a Terraform validation error before the request ever reached the ACK API. The web console offered AL4 as an OS choice, so I knew the OS was supported by ACK end-to-end. The only blocker was  the provider's allow-list.

To work around it, I ended up dropping `image_type` entirely and resolving an image ID through `data "alicloud_images"` with a regex against the AL4 ARM64 image name family:

```hcl
data "alicloud_images" "alinux4_arm64" {
  most_recent = true
  owners      = "system"
  name_regex  = "^aliyun_4_arm64_20G_alibase_.*"
}

resource "alicloud_cs_kubernetes_node_pool" "workers" {
  # ...
  instance_types = ["ecs.g8y.large"]
  image_id       = data.alicloud_images.alinux4_arm64.images[0].id
  # image_type intentionally omitted because "AliyunLinux4Arm64" is rejected
}
```

That works, but it's a fair amount of friction (and a fragile name regex) for what should be a one-line `image_type` assignment. It also took me a while to figure out, as the docs don't mention the workaround, and the natural value to try (`AliyunLinux4Arm64`, by analogy with the existing `AliyunLinux3Arm64`) gets a validation error rather than a hint.

## Desired state

Users can write `image_type = "AliyunLinux4"`, `"AliyunLinux4Arm64"`, or `"AliyunLinux4Security"` and have it pass schema validation, mirroring the v3 family. The docs accurately list every value the schema accepts.

## Changes

1. **`alicloud/resource_alicloud_cs_kubernetes_node_pool.go`**: Extend the `image_type` `StringInSlice` allow-list with `AliyunLinux4`, `AliyunLinux4Arm64`, `AliyunLinux4Security`. Grouped next to the existing `AliyunLinux4ContainerOptimized` entry.
2. **`website/docs/r/cs_kubernetes_node_pool.html.markdown`**: Add bullets for the three new v4 values, and incidentally fix the pre-existing parity gap for `AliyunLinux Qboot` and `AliyunLinuxSecurity`.
3. **`website/docs/d/cs_kubernetes_node_pools.html.markdown`**: Sync the inline `image_type` description to list all currently accepted values (was missing five).

The deprecated `platform` field's narrower allow-list (line 645) is intentionally not touched.

## What I'd like maintainers to verify

- **`AliyunLinux4Arm64`**: I have seen that this image family is supported by ACK end-to-end (my node pool runs on it via `image_id`), but I haven't actually round-tripped the literal string `"AliyunLinux4Arm64"` through the `image_type` API parameter. Confidence is high but not 100%.
- **`AliyunLinux4`** (plain x86) and **`AliyunLinux4Security`**: Added by analogy with the v3 family. I haven't validated that the API accepts these literal strings - please confirm or correct based on the actual CreateClusterNodePool enum. Happy to drop either if they shouldn't be there yet.

The public API reference page does not currently list any v4 image types (including the already-shipping `AliyunLinux4ContainerOptimized`), so I couldn't cross-check against it.

## Test plan

- `make fmtcheck` - clean.
- `go vet ./alicloud/` - no new diagnostics introduced by this change.
- `go build ./...`  - succeeds.
- No new acceptance tests added, matching the precedent of the previous `image_type` additions (`AliyunLinux4ContainerOptimized`, `Custom`, `Ubuntu`), which were validation-only changes.


alicloud/resource_alicloud_cs_kubernetes_node_pool.go
```diff
@@ -190,7 +190,7 @@ func resourceAliCloudAckNodepool() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Computed:     true,
-				ValidateFunc: StringInSlice([]string{"AliyunLinux", "AliyunLinux3", "AliyunLinux3Arm64", "AliyunLinuxUEFI", "CentOS", "Windows", "WindowsCore", "AliyunLinux Qboot", "ContainerOS", "AliyunLinuxSecurity", "Ubuntu", "AliyunLinux3ContainerOptimized", "Custom", "AliyunLinux4ContainerOptimized"}, false),
+				ValidateFunc: StringInSlice([]string{"AliyunLinux", "AliyunLinux3", "AliyunLinux3Arm64", "AliyunLinuxUEFI", "CentOS", "Windows", "WindowsCore", "AliyunLinux Qboot", "ContainerOS", "AliyunLinuxSecurity", "Ubuntu", "AliyunLinux3ContainerOptimized", "Custom", "AliyunLinux4ContainerOptimized", "AliyunLinux4", "AliyunLinux4Arm64", "AliyunLinux4Security"}, false),
```
xuzhang3: `AliyunLinux Qboot` is not supported, should removed.

xuzhang3 (changes requested):

alicloud/resource_alicloud_cs_kubernetes_node_pool.go
grammeaway: @xuzhang3 my latest commit should clean the system of Qboot references :v:

PR #9749, opened 2026-05-22
New Resource: alicloud_polardb_aicluster

kittenshow98:
New Resource: alicloud_polardb_aicluster

### DIFF
--- a/alicloud/provider.go
+++ b/alicloud/provider.go
@@ -1547,6 +1547,7 @@ func Provider() terraform.ResourceProvider {
 			"alicloud_polardb_application":                                   resourceAlicloudPolarDBApplication(),
 			"alicloud_polardb_application_endpoint":                          resourceAlicloudPolarDBApplicationEndpoint(),
 			"alicloud_polardb_batch_task":                                    resourceAlicloudPolarDBBatchTask(),
+			"alicloud_polardb_aicluster":                                     resourceAlicloudPolarDBAICluster(),
 			"alicloud_hbase_instance":                                        resourceAlicloudHBaseInstance(),
 			"alicloud_market_order":                                          resourceAlicloudMarketOrder(),
 			"alicloud_adb_cluster":                                           resourceAliCloudAdbDbCluster(),
--- a/alicloud/resource_alicloud_polardb_aicluster.go
+++ b/alicloud/resource_alicloud_polardb_aicluster.go
@@ -1,0 +1,349 @@
+package alicloud
+
+import (
+	"fmt"
+	"strings"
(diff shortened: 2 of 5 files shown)


xuzhang3:
### Code review

Found 4 issues:

1. Typo `ModeName` → `ModelName` silently drops `model_name` on every Create. The Read function reads `attribute["ModelName"]` (line 292) so the intended API field is `ModelName`; as written, the user's `model_name` is never sent to `CreateAIDBCluster`.

https://github.com/aliyun/terraform-provider-alicloud/blob/5ce3fbf3a3a0840f90ff3ca7017691d286cccbb4/alicloud/resource_alicloud_polardb_aicluster.go#L196-L204

2. `db_cluster_id` (ForceNew) is overwritten with the AI cluster's own id on Read. `DescribePolarDBAIClusterAttribute` is queried by `d.Id()` (the AI cluster id), so `attribute["DBClusterId"]` is the AI cluster's own id — already assigned to `ai_cluster_id` at line 265-267. Reusing the same key here corrupts the user-supplied associated cluster id; every subsequent plan will diff and want to recreate the resource. Use the correct response key for the associated cluster id, or drop the `d.Set` if Describe does not return it.

https://github.com/aliyun/terraform-provider-alicloud/blob/5ce3fbf3a3a0840f90ff3ca7017691d286cccbb4/alicloud/resource_alicloud_polardb_aicluster.go#L301-L308

3. Six `Optional` fields are not `ForceNew` and the resource has no `Update` — any post-create change errors with `"alicloud_polardb_aicluster" doesn't support update`. Affected fields: `auto_renew`, `period`, `used_time` (linked below), plus `inference_engine` (L111-L116), `auto_use_coupon` (L124-L129), `promotion_code` (L130-L134). Add `ForceNew: true` or implement Update.

https://github.com/aliyun/terraform-provider-alicloud/blob/5ce3fbf3a3a0840f90ff3ca7017691d286cccbb4/alicloud/resource_alicloud_polardb_aicluster.go#L55-L75

4. `auto_use_coupon = false` cannot be expressed — `d.GetOk` returns `ok=false` for the zero value of a bool, so an explicit `false` is dropped and the API still sees the schema default `true`. Use `d.GetOkExists` (or read the raw value) so booleans round-trip.

https://github.com/aliyun/terraform-provider-alicloud/blob/5ce3fbf3a3a0840f90ff3ca7017691d286cccbb4/alicloud/resource_alicloud_polardb_aicluster.go#L208-L216

kittenshow98:
image

kittenshow98:
image

PR #9834, opened 2026-06-09
resource/alicloud_ots_table: Fix tablestore data plane ignoring HTTP(S)_PROXY under cross-border proxy

ChenHanZhang:
## Problem
OTS data-plane operations (table/secondary index/search index/tunnel) ignore the proxy and force a direct connection. Where the region is reachable only via a proxy (e.g. cross-border CI runner), control-plane and OSS state work but any OTS data-plane call fails with `net.OpError` on Post. Refs Aone Req 82959751, bug 82899631.

## Root cause
`WithOtsClient` (control plane) sets the proxy via `SetHttpProxy/SetHttpsProxy`, but `WithTableStoreClient`/`WithTableStoreTunnelClient` build the SDK client with `NewDefaultTableStoreConfig()`/`DefaultTunnelConfig`, whose transport has no Proxy, so the proxy is ignored.

## Fix
Add `getOtsProxy()` (HTTP(S)_PROXY + NO_PROXY, same logic as the control plane) and set the proxy on the TableStore/tunnel transport. Direct-connect path is untouched (proxy nil -> SDK default).

## Test
- New unit tests: getOtsProxy + table/tunnel transport, proxy vs direct vs NO_PROXY (all pass)
- Remote ACC: TestAccAliCloudOtsTable_basic PASS (direct path not broken)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/alicloud/connectivity/client.go
+++ b/alicloud/connectivity/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


ChenHanZhang:
Closing; will resubmit a clean PR with an added OTS integration test and rewritten commit history.

PR #9876, opened 2026-06-23
New Resource: alicloud_cloud_monitor_service_metric_alarm_rule; data-source/alicloud_cloud_monitor_service_metric_alarm_rules: regenerate from CMS IDL

LCLSpring:
## Summary

- Release the new `alicloud_cloud_monitor_service_metric_alarm_rule` resource and rewrite the existing `alicloud_cloud_monitor_service_metric_alarm_rules` data source on top of CMS `DescribeMetricRuleList` / `PutResourceMetricRule` / `PutMetricRuleTargets` / `DeleteMetricRules` / `EnableMetricRules` / `DisableMetricRules` / `DescribeMetricRuleTargets`.
- After confirming with the CMS team (Aone 83358430) that `PutResourceMetricRule` is a full-replace PUT, the `Update` path now sends the complete payload on every call, including the `composite_expression.expression_list` array that the generator template had been dropping. This was the root cause of several Apply-time fields (effective_interval, no_data_policy, email_subject, webhook, ...) being reset to defaults after the second PUT.
- `Read` no longer surfaces the synthetic `Escalations.Critical` rows that CMS injects for PROMETHEUS-style rules, and refreshes `targets` only when `DescribeMetricRuleTargets` returns entries, so resources that never managed targets do not pick up stale rows during import/refresh.
- Schema cleanups: drop `ForceNew` from `composite_expression` / `escalations` (CMS PUT updates them in place), and mark `email_subject` and `targets` as `Computed` alongside `Optional` so the server-side defaults stop showing as plan diffs.
- Data source bug fixes: stop forcing `metric_alarm_rule_id` / `metric_name` / `namespace` / `rule_name` / `status` defaults into `DescribeMetricRuleList`, replace the residual JSONPath key when filtering by ids, and switch the test fixture `status` filter to a bool literal so the data source schema validation passes.

## Test plan

- [x] Remote ACC: 8 resource cases (`TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12863..12870`) all pass
- [x] Remote ACC: data source case (`TestAccAlicloudCloudMonitorServiceMetricAlarmRuleDataSource`) passes

```
=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12863
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12863 (10.34s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12864
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12864 (9.21s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12865
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12865 (8.74s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12866
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12866 (16.41s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12867
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12867 (10.85s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12868
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12868 (8.79s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12869
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12869 (12.13s)

=== RUN   TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12870
--- PASS: TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12870 (9.42s)

=== RUN   TestAccAlicloudCloudMonitorServiceMetricAlarmRuleDataSource
--- PASS: TestAccAlicloudCloudMonitorServiceMetricAlarmRuleDataSource (147.32s)
```

### DIFF
--- a/website/docs/d/cloud_monitor_service_metric_alarm_rules.html.markdown
+++ b/website/docs/d/cloud_monitor_service_metric_alarm_rules.html.markdown
@@ -3,18 +3,16 @@ subcategory: "Cloud Monitor Service"
 layout: "alicloud"
 page_title: "Alicloud: alicloud_cloud_monitor_service_metric_alarm_rules"
 description: |-
-  Provides a list of Cloud Monitor Service Metric Alarm Rules to the user.
+  Provides a list of Cloud Monitor Service Metric Alarm Rule owned by an Alibaba Cloud account.
 ---
 
 # alicloud_cloud_monitor_service_metric_alarm_rules
 
-This data source provides the Cloud Monitor Service Metric Alarm Rules of the current Alibaba Cloud user.
+This data source provides Cloud Monitor Service Metric Alarm Rule available to the user.[What is Metric Alarm Rule](https://next.api.alibabacloud.com/document/Cms/2019-01-01/PutResourceMetricRule)
 
 -> **NOTE:** Available since v1.256.0.
 
 ## Example Usage
-
-Basic Usage
 
 ```terraform
 variable "name" {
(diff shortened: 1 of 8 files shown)


api-tool-agent:
**Scope update: the alarm-callback `targets` block has been removed from this resource.**

The CMS team confirmed (internal ticket Aone 83358430) that:
- `PutMetricRuleTargets` silently discards unsupported writes — it returns success without persisting `JsonParams`;
- `DescribeMetricRuleTargets` never returns that field, so the value can never round-trip;
- the Target APIs are an independent object that had been mistakenly attached to this resource's model.

Shipping the block would have meant publishing a field that silently loses data. Managing MNS callback targets remains fully available through the existing `alicloud_cms_alarm` resource, so no capability is lost. Since this resource has never been released, the removal is not a breaking change.

**Latest remote ACC results (after the removal, on the current head):** all 24 resource cases pass (`TestAccAliCloudCloudMonitorServiceMetricAlarmRule_basic12863..12870` plus their `_twin` and `_raw` variants) and `TestAccAlicloudCloudMonitorServiceMetricAlarmRuleDataSource` passes — 25/25 green.

website/docs/d/cloud_monitor_service_metric_alarm_rules.html.markdown
```diff
@@ -2,20 +2,19 @@
...
+  Provides a list of Cloud Monitor Service Metric Alarm Rule owned by an Alibaba Cloud account.
 ---
 
 # alicloud_cloud_monitor_service_metric_alarm_rules
 
-This data source provides the Cloud Monitor Service Metric Alarm Rules of the current Alibaba Cloud user.
+This data source provides Cloud Monitor Service Metric Alarm Rule available to the user.[What is Metric Alarm Rule](https://next.api.alibabacloud.com/document/Cms/2019-01-01/PutResourceMetricRule)
 
--> **NOTE:** Available since v1.256.0.
+-> **NOTE:** Available since v1.283.0.
```
LCLSpring: 改这个是不是不太合适？这不是新增资源。

api-tool-agent: Good catch — fixed in the latest push. The generator regenerated this existing data source's doc and wrongly bumped the version note; restored it to `Available since v1.256.0` (the version this data source was originally released in), and also dropped the stray `sidebar_current` line the generator added. The new resource's doc note was corrected to `v1.285.0` to match the upcoming release.

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (commented):
approved

PR #9898, opened 2026-06-29
resource/alicloud_db_backup_policy: Added enable_pitr_protection and inc_backup_interval, optimized backup_interval and backup_retention_period

chaitiansheng0524:
- Added `enable_pitr_protection` (TypeBool, MySQL only, LogBackupPolicy mode)
- Added `inc_backup_interval` (TypeInt, DataBackupPolicy mode)
- Fixed `backup_interval` default value bug: no longer sends -1 when not set
- Fixed `backup_retention_period` bug: removed != 7 check that prevented setting value back to default 7

### DIFF
--- a/alicloud/resource_alicloud_db_backup_policy.go
+++ b/alicloud/resource_alicloud_db_backup_policy.go
@@ -177,6 +177,17 @@ func resourceAlicloudDBBackupPolicy() *schema.Resource {
 				Computed:     true,
 				ValidateFunc: validation.StringInSlice([]string{"-1", "15", "30", "60", "120", "180", "240", "360", "480", "720"}, false),
 			},
+			"enable_pitr_protection": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+			},
+			"inc_backup_interval": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validation.IntInSlice([]int{60, 120, 240, 360, 720}),
+			},
 			"backup_priority": {
 				Type:         schema.TypeInt,
 				Optional:     true,
@@ -263,6 +274,14 @@ func resourceAlicloudDBBackupPolicyRead(d *schema.ResourceData, meta interface{}
 	d.Set("archive_backup_retention_period", formatInt(object["ArchiveBackupRetentionPeriod"]))
 	d.Set("archive_backup_keep_count", formatInt(object["ArchiveBackupKeepCount"]))
 	d.Set("archive_backup_keep_policy", object["ArchiveBackupKeepPolicy"])
+	if instance["Engine"] == "MySQL" && instance["DBInstanceStorageType"] == "local_ssd" && object["EnableIncrementDataBackup"].(bool) {
+		d.Set("inc_backup_interval", object["IncBackupInterval"])
+	}
+	if instance["Engine"] == "MySQL" && object["EnableBackupLog"] == "1" {
+		if v, ok := object["EnablePitrProtection"]; ok && v != nil {
+			d.Set("enable_pitr_protection", v.(bool))
+		}
+	}
 	return nil
 }
 
@@ -274,8 +293,8 @@ func resourceAlicloudDBBackupPolicyUpdate(d *schema.ResourceData, meta interface
 
 	if d.HasChanges("backup_period", "backup_time", "retention_period", "preferred_backup_period", "preferred_backup_time",
 		"backup_retention_period", "compress_type", "log_backup_frequency", "archive_backup_retention_period", "archive_backup_keep_count",
-		"archive_backup_keep_policy", "released_keep_policy", "category", "backup_interval", "backup_priority", "backup_method",
-		"enable_increment_data_backup") {
+		"archive_backup_keep_policy", "released_keep_policy", "category", "backup_interval", "inc_backup_interval", "backup_priority",
+		"backup_method", "enable_increment_data_backup", "enable_pitr_protection") {
 		updateForData = true
 	}
 
--- a/alicloud/resource_alicloud_db_backup_policy_test.go
+++ b/alicloud/resource_alicloud_db_backup_policy_test.go
@@ -66,6 +66,16 @@ func TestAccAliCloudRdsDBBackupPolicyMySql(t *testing.T) {
 			},
 			{
 				Config: testAccConfig(map[string]interface{}{
+					"enable_pitr_protection": true,
+				}),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheck(map[string]string{
+						"enable_pitr_protection": "true",
+					}),
+				),
+			},
+			{
+				Config: testAccConfig(map[string]interface{}{
 					"backup_retention_period": "600",
 					"retention_period":        "600",
 				}),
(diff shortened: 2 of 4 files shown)


xuzhang3:
```log
=== 存量测试（eu-central-1）===
--- PASS: TestAccAliCloudRdsDBBackupPolicySQLServer (789.79s)
--- PASS: TestAccAliCloudRdsDBBackupPolicySQLServerAlwaysOn (1046.61s)
--- PASS: TestAccAliCloudRdsDBBackupPolicyMariaDB (2546.49s)

=== PostgreSQL（cn-beijing）===
--- PASS: TestAccAliCloudRdsDBBackupPolicyPostgreSQL (1777.10s)

=== MySQL（cn-beijing）===
--- PASS: TestAccAliCloudRdsDBBackupPolicyMySql (1330.28s)

全部 5/5 ✅
```

xuzhang3 (commented):
I ran the full `TestAccAliCloudRdsDBBackupPolicy*` suite against this branch plus two additional changes in `alicloud/service_alicloud_rds.go`, and I think both are needed before `inc_backup_interval` can be trusted. Details and a suggested diff below.

## Why the current branch needs a provider-side change

**1. `inc_backup_interval` is a false green on MySQL local disk.**

`Read` only writes the attribute when incremental backup is reported as enabled:

```go
// alicloud/resource_alicloud_db_backup_policy.go
if instance["Engine"] == "MySQL" && instance["DBInstanceStorageType"] == "local_ssd" && object["EnableIncrementDataBackup"].(bool) {
	d.Set("inc_backup_interval", object["IncBackupInterval"])
}
```

But `ModifyDBBackupPolicy` only sends `EnableIncrementDataBackup` on the `SQLServer` + non-local-disk path — nothing sends it for MySQL local disk. So on a fresh local-disk instance the API keeps `EnableIncrementDataBackup=false`, `Describe` returns false, the gate never fires, and `inc_backup_interval` is never written to state. Because the attribute is `Optional + Computed`, an unset attribute compares equal to whatever the check expects, so a test asserting `inc_backup_interval` passes without the value ever having round-tripped. I confirmed this directly: removing that one send turns the local-disk case from PASS into a FAIL on the step that checks `60`.

**2. `Category` breaks every update after create on a local-disk instance.**

`Category` (秒级备份 / flash backup) goes into the `DataBackupPolicy` request unconditionally, and the schema field is `Optional + Computed` with `Read` writing it back. So:

- create — config and state both empty, `Category: ""` is sent, backend treats it as unset, **succeeds**
- `Read` stores the queried value
- every later apply — `GetOk` returns that state value, `Category` is now sent for real, and a local-disk instance rejects it with a 500 `InternalError`

The non-obvious part: the rejection is not specific to `Flash`. The backend fails the assertion because the instance does not support the flash feature *at all*, so `Standard` is rejected too — the empty string is the only value that gets through. `if category != ""` therefore does not fix it, because under SDK v1 a Computed attribute's `GetOk` cannot distinguish "set in config" from "populated into state by Read".

This is why a two-apply local-disk test (create at `60`, update to `120`) cannot pass on the current branch — the second apply always 500s.

## Suggested diff

```diff
 			"BackupPolicyMode":      "DataBackupPolicy",
 			"SourceIp":              s.client.SourceIp,
 			"ReleasedKeepPolicy":    releasedKeepPolicy,
-			"Category":              category,
+		}
+		// MySQL local disk instances do not support the flash backup feature and reject Category
+		// whatever its value. Category is Computed, so once Read has stored the queried value it
+		// would otherwise be sent back on every subsequent update
+		if !(instance["Engine"] == "MySQL" && instance["DBInstanceStorageType"] == "local_ssd") {
+			request["Category"] = category
 		}
 		if instance["Engine"] == "MySQL" && instance["DBInstanceStorageType"] == "local_ssd" {
+			// Incremental backup applies to SQL Server cloud disk and MySQL local disk instances,
+			// and IncBackupInterval only takes effect once it is enabled
+			request["EnableIncrementDataBackup"] = enableIncrementDataBackup
 			if v, ok := d.GetOk("inc_backup_interval"); ok {
 				request["IncBackupInterval"] = v.(int)
 			}
 		}
```

Two caveats on that diff, both worth a look before it lands:

- The `Category` gate as written is narrower than the constraint it encodes. The limitation is a local-disk one, not a MySQL one, so `instance["DBInstanceStorageType"] != "local_ssd"` across all engines is probably the correct predicate — SQL Server / PostgreSQL / MariaDB on local disk should hit the same 500 on a second apply. I have not covered that, since there is no non-MySQL local-disk case in the suite; given the blast radius it may be better as its own change.
- `IncBackupInterval` is still sent whenever it is set in config, including when `enable_increment_data_backup` is `false`. Both local-disk applies in my run had it enabled, so that combination is untested.

Also worth noting the documented constraint for `inc_backup_interval` in `website/docs/r/db_backup_policy.html.markdown` does not match what the API accepts — incremental backup applies to SQL Server cloud disk and MySQL **local** disk instances, so the doc text needs a pass too. I left the docs alone here rather than widen this PR.

## Acceptance test results

Full suite, this branch plus the diff above, all six cases:

```
PASS: TestAccAliCloudRdsDBBackupPolicyMySql (1276.05s)
PASS: TestAccAliCloudRdsDBBackupPolicyMySqlLocalDisk (846.36s)
PASS: TestAccAliCloudRdsDBBackupPolicyMariaDB (2146.12s)
PASS: TestAccAliCloudRdsDBBackupPolicySQLServer (829.67s)
PASS: TestAccAliCloudRdsDBBackupPolicySQLServerAlwaysOn (1174.22s)
PASS: TestAccAliCloudRdsDBBackupPolicyPostgreSQL (1965.28s)

6 passed / 0 failed / 0 skipped (6 total)
```

Verified from the debug traffic rather than the tally alone:

- of the `DataBackupPolicy` requests in the run, every one against a cloud-disk instance carries `Category` (empty on create, then `Flash` or `Standard`), and **both** against the local-disk instance have `Category` absent — the gate does what it claims
- the local-disk pair sends `IncBackupInterval` `60` then `120` with `EnableIncrementDataBackup=true`, and the `ModifyBackupPolicy` **response** echoes `IncBackupInterval:120` — so the second apply, the one that used to 500, now succeeds
- `DescribeBackupPolicy` responses return `IncBackupInterval` `60` and `120`, so the read path is API-sourced rather than echoing state
- zero `InternalError` and zero flash-rejection responses anywhere in the run
- `EnablePitrProtection` appears in both requests and `Describe` responses, true and false

For reference, the same suite on this branch **without** the diff: 5 passed / 1 failed, the failure being `TestAccAliCloudRdsDBBackupPolicyMySqlLocalDisk` at the step asserting `inc_backup_interval = 60`.

I also reshaped the local-disk coverage into a separate `TestAccAliCloudRdsDBBackupPolicyMySqlLocalDisk` case (`local_ssd`, create with `enable_increment_data_backup = true` + `inc_backup_interval = 60`, update to `120`, then `ImportStateVerify`) and restored `TestAccAliCloudRdsDBBackupPolicyMySql` to its cloud-disk fixture, since `inc_backup_interval` cannot apply there. Happy to hand that over as a patch if useful.

xuzhang3:
Heads-up: I've applied the diff from my review above directly to this branch rather than leaving it for you to pick up.

I amended it into your existing commit instead of adding a second one, because this repo's CI caps a PR at one commit. Your authorship is preserved — the commit is still authored by you, only the committer and the SHA changed (`de152cb2` → `c935e9b2`).

**Your local branch has diverged as a result.** To sync without creating a merge commit:

```
git fetch origin
git reset --hard origin/feature_alicloud_db_backup_policy_enable_pitr_protection
```

If you have unpushed local work on this branch, stash or branch it off before that reset.

What the amend added, on top of what you already had:

- `service_alicloud_rds.go` — `Category` is now gated so it is not sent for MySQL local-disk instances, and `EnableIncrementDataBackup` is now sent on the MySQL local-disk path. Rationale and the failure mode for each are in my review comment above.
- `resource_alicloud_db_backup_policy_test.go` — `TestAccAliCloudRdsDBBackupPolicyMySql` is back on its cloud-disk fixture with the `inc_backup_interval` steps removed, since that attribute does not apply there, and the local-disk coverage moved into a new `TestAccAliCloudRdsDBBackupPolicyMySqlLocalDisk` case (create at `60`, update to `120`, then `ImportStateVerify`).

Full suite on the amended branch: 6 passed / 0 failed / 0 skipped, per-case timings in my review comment. Please sanity-check the two caveats I listed there — in particular the `Category` gate is written MySQL-specific while the underlying limitation looks local-disk-wide, and I could not cover a non-MySQL local-disk instance.

PR #9916, opened 2026-07-02
resource/alicloud_esa_routine: , resource/alicloud_esa_routine_related_record: , resource/alicloud_esa_routine_route: retry LockFailed on concurrent writes

api-tool-agent:
## Problem

Concurrently creating multiple `alicloud_esa_routine_route` resources under the same `SiteId` + `RoutineName` returns `LockFailed` from the ESA API, and terraform apply fails. `LockFailed` is ESA's server-side optimistic-locking error code for concurrent writes to the same routine configuration; it always co-occurs with `Site.ServiceBusy` and `TooManyRequests`, and is a normal transient concurrency-control response that clients are expected to retry with backoff.

## Root cause

The provider's `NeedRetry(err)` helper only covers network / timeout / throttling / 5xx / `ServiceUnavailable`. `LockFailed` is not in the set, so the retry loop treats it as `NonRetryableError` and bails out immediately.

## Fix

Wrap each RPC retry site with `IsExpectedErrors(err, []string{"Site.ServiceBusy", "TooManyRequests", "LockFailed"}) || NeedRetry(err)`, using the exact pattern already established by 14+ other alicloud ESA resources (`esa_image_transform`, `esa_cache_rule`, ...).

### RPC retry sites patched (4 files, 8 insertions, 8 deletions)

| File | Function |
|---|---|
| `alicloud/resource_alicloud_esa_routine.go` | Create, Delete |
| `alicloud/resource_alicloud_esa_routine_related_record.go` | Create, Delete |
| `alicloud/resource_alicloud_esa_routine_route.go` | Create, Update, Delete |
| `alicloud/service_alicloud_esa_v2.go` | `DescribeEsaRoutineRoute` Read |

## Regression test

`TestAccAliCloudEsaRoutineRoute_lockRetry` — declares 8 `alicloud_esa_routine_route` resources under one Site + Routine and drives Create → Update → Destroy in parallel to stress the retry sites end-to-end. Passes locally after the fix.

## Notes

- No `MutexKV` / client-side lock added — the concurrency guard is enforced server-side by ESA; the provider only needs to translate the transient error into a retryable one.
- No import, signature, or call-site changes. Pure `if`-condition edit; risk is minimal.
- The affected files carry the `// This file is generated automatically` header; the generator has been missing this retry pattern for routine-family resources.

### DIFF
--- a/alicloud/resource_alicloud_esa_routine.go
+++ b/alicloud/resource_alicloud_esa_routine.go
@@ -63,7 +63,7 @@ func resourceAliCloudEsaRoutineCreate(d *schema.ResourceData, meta interface{}) 
 	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
 		response, err = client.RpcPost("ESA", "2024-09-10", action, query, request, true)
 		if err != nil {
-			if NeedRetry(err) {
+			if IsExpectedErrors(err, []string{"Site.ServiceBusy", "TooManyRequests", "LockFailed"}) || NeedRetry(err) {
 				wait()
 				return resource.RetryableError(err)
 			}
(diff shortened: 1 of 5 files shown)


api-tool-agent:
recheck

ChenHanZhang (approved):

PR #9920, opened 2026-07-02
resource/alicloud_eflo_cluster: support node group provisioning fields and computed node_group_ids

amontalban:
### Why

The `alicloud_eflo_cluster` resource cannot fully provision Lingjun (EFLO) GPU node groups today. Creating a node group for GPU machines (e.g. B300 / `efg2.GN9D`) requires:

- group-level node authentication (`key_pair_name` / `login_password`),
- an RDMA bond count (`bond_num`),
- a cloud system disk — so the backend boots from a cloud disk instead of building a whitelist-restricted local disk group,
- the high-performance network zone (`hpn_zone`),
- optional file-system mounting (`file_system_mount_enabled`).

None of these are exposed, so node groups created through Terraform cannot provision these machine types. There is also no way to obtain the generated node group IDs, which are required to bind an ACK `type = "lingjun"` node pool to the cluster via `eflo_node_group.group_id`.

### What are you changing?

Add the following optional arguments to `node_groups`:

- `key_pair_name` — key pair used to log on to the nodes
- `login_password` — password used to log on to the nodes (sensitive)
- `bond_num` — number of RDMA network bonds
- `file_system_mount_enabled` — enable file system mounting
- `hpn_zone` — high-performance network zone
- `system_disk` block — `size`, `performance_level`, `category`

Add a computed top-level `node_group_ids` (map of node group name → ID), populated from `ListNodeGroups` in the read, so an ACK Lingjun node pool can reference the group.

The resource documentation is updated accordingly. `gofmt` is clean and `go build ./alicloud/` passes. All changes are additive and backward-compatible (new arguments are Optional; `node_group_ids` is Computed).

---

## Coverage & remaining SDK options

`alicloud_eflo_cluster` dispatches `CreateCluster` through the generic `client.RpcPost("eflo-controller", "2022-12-15", ...)` path with a hand-built `map[string]interface{}` request (`node_groups` is JSON-marshaled into `request["NodeGroups"]`), not a typed request struct. Each new attribute is therefore added as a schema field plus a matching entry in the Create request map (see the `node_groups` loop), and `node_group_ids` is populated on Read from `ListNodeGroups`.

**Fields this PR adds to `node_groups`:** `key_pair_name`, `login_password` (Sensitive), `file_system_mount_enabled`, `system_disk` block (`size`, `performance_level`, `category`), plus `bond_num` and node-group-level `hpn_zone` (RDMA bonding / HPN zone). It also adds the computed top-level `node_group_ids` map (node-group name → generated id) so an ACK `type="lingjun"` node pool can bind to the cluster.

### SDK model additions not yet exposed

Because the provider hand-builds the request map rather than consuming the typed [`eflo-controller-20221215`](https://github.com/alibabacloud-go/eflo-controller-20221215) Go SDK, it does not automatically track additions to the SDK's `CreateClusterRequestNodeGroups` model. The node-group fields below were added to that SDK model in the releases shown and are all still present at the latest SDK release (v3.0.7, `client/create_cluster_request_model.go`), but none are surfaced by the provider's `resource_alicloud_eflo_cluster.go` schema. They are out of scope for this PR (future work):

| SDK model field | SDK struct | Added to SDK in | Provider status |
| --- | --- | --- | --- |
| `Nodes[].DataDisk` | `CreateClusterRequestNodeGroupsNodes` (`DataDisk []*CreateClusterRequestNodeGroupsNodesDataDisk`) | [v2.6.3](https://github.com/alibabacloud-go/eflo-controller-20221215/releases/tag/v2.6.3) | Still not exposed |
| `VirtualGpuEnabled` | `CreateClusterRequestNodeGroups` (`VirtualGpuEnabled *bool`) | [v2.6.4](https://github.com/alibabacloud-go/eflo-controller-20221215/releases/tag/v2.6.4) | Still not exposed |
| `HyperNodes` | `CreateClusterRequestNodeGroups` (`HyperNodes []*CreateClusterRequestNodeGroupsHyperNodes`) | [v2.8.3](https://github.com/alibabacloud-go/eflo-controller-20221215/releases/tag/v2.8.3) | Still not exposed |
| `RamRoleName` | `CreateClusterRequestNodeGroups` (`RamRoleName *string`) | [v3.0.6](https://github.com/alibabacloud-go/eflo-controller-20221215/releases/tag/v3.0.6) | Still not exposed |

## Docs & tests

- Documents the new arguments, and marks the create-only blocks (`networks`, `node_groups`, `components`) as **immutable after resource creation** — the resource has no update path for them (they sit in `ImportStateVerifyIgnore`), so this reflects real behavior.
- Extends `TestAccAliCloudEfloCluster_basic10311` to cover the new `node_groups` attributes (`key_pair_name`, `login_password`, `bond_num`, `file_system_mount_enabled`, `hpn_zone`, `system_disk`) plus `nodes`, a `components` block, and the computed `node_group_ids`. Placeholder values (e.g. on `efg1.nvga1n` in `cn-wulanchabu`) can be adjusted for a real acceptance run.

### DIFF
--- a/alicloud/resource_alicloud_eflo_cluster.go
+++ b/alicloud/resource_alicloud_eflo_cluster.go
@@ -312,6 +312,48 @@ func resourceAliCloudEfloCluster() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+						"key_pair_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"login_password": {
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
+						},
+						"bond_num": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"file_system_mount_enabled": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+						"hpn_zone": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"system_disk": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"size": {
+										Type:     schema.TypeInt,
+										Optional: true,
+									},
+									"performance_level": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"category": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+								},
+							},
+						},
 						"nodes": {
 							Type:     schema.TypeList,
 							Optional: true,
@@ -355,6 +397,11 @@ func resourceAliCloudEfloCluster() *schema.Resource {
 			"status": {
 				Type:     schema.TypeString,
 				Computed: true,
+			},
+			"node_group_ids": {
+				Type:     schema.TypeMap,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"tags": tagsSchema(),
 		},
@@ -578,6 +625,35 @@ func resourceAliCloudEfloClusterCreate(d *schema.ResourceData, meta interface{})
 			dataLoop8Map["NodeGroupName"] = dataLoop8Tmp["node_group_name"]
 			dataLoop8Map["MachineType"] = dataLoop8Tmp["machine_type"]
 			dataLoop8Map["ImageId"] = dataLoop8Tmp["image_id"]
+			if v, ok := dataLoop8Tmp["key_pair_name"]; ok && v.(string) != "" {
+				dataLoop8Map["KeyPairName"] = v
+			}
+			if v, ok := dataLoop8Tmp["login_password"]; ok && v.(string) != "" {
+				dataLoop8Map["LoginPassword"] = v
+			}
+			if v, ok := dataLoop8Tmp["hpn_zone"]; ok && v.(string) != "" {
+				dataLoop8Map["HpnZone"] = v
+			}
+			if v, ok := dataLoop8Tmp["bond_num"]; ok && v.(int) > 0 {
+				dataLoop8Map["BondNum"] = v
+			}
+			if v, ok := dataLoop8Tmp["file_system_mount_enabled"]; ok {
+				dataLoop8Map["FileSystemMountEnabled"] = v
+			}
+			if v, ok := dataLoop8Tmp["system_disk"].([]interface{}); ok && len(v) > 0 && v[0] != nil {
+				sdm := v[0].(map[string]interface{})
+				systemDisk := make(map[string]interface{})
+				if x, ok := sdm["size"]; ok && x.(int) > 0 {
+					systemDisk["Size"] = x
+				}
+				if x, ok := sdm["performance_level"]; ok && x.(string) != "" {
+					systemDisk["PerformanceLevel"] = x
+				}
+				if x, ok := sdm["category"]; ok && x.(string) != "" {
+					systemDisk["Category"] = x
+				}
+				dataLoop8Map["SystemDisk"] = systemDisk
+			}
 			nodeGroupsMapsArray = append(nodeGroupsMapsArray, dataLoop8Map)
 		}
 		nodeGroupsMapsJson, err := json.Marshal(nodeGroupsMapsArray)
@@ -699,6 +775,53 @@ func resourceAliCloudEfloClusterRead(d *schema.ResourceData, meta interface{}) e
 
 	tagsMaps, _ := jsonpath.Get("$.TagResources.TagResource", objectRaw)
 	d.Set("tags", tagsToMap(tagsMaps))
+
+	nodeGroupIds := make(map[string]interface{})
+	listAction := "ListNodeGroups"
+	listRequest := map[string]interface{}{"ClusterId": d.Id(), "RegionId": client.RegionId, "MaxResults": PageSizeLarge}
+	listQuery := make(map[string]interface{})
+	for {
+		var listResponse map[string]interface{}
+		wait := incrementalWait(3*time.Second, 5*time.Second)
+		err = resource.Retry(1*time.Minute, func() *resource.RetryError {
+			listResponse, err = client.RpcPost("eflo-controller", "2022-12-15", listAction, listQuery, listRequest, true)
+			if err != nil {
+				if NeedRetry(err) {
+					wait()
+					return resource.RetryableError(err)
+				}
+				return resource.NonRetryableError(err)
+			}
+			return nil
+		})
+		addDebug(listAction, listResponse, listRequest)
+		if err != nil {
+			return WrapErrorf(err, DefaultErrorMsg, d.Id(), listAction, AlibabaCloudSdkGoERROR)
+		}
+
+		groupsRaw, _ := jsonpath.Get("$.Groups", listResponse)
+		if groups, ok := groupsRaw.([]interface{}); ok {
+			for _, g := range groups {
+				gm, ok := g.(map[string]interface{})
+				if !ok {
+					log.Printf("[WARN] Resource alicloud_eflo_cluster ListNodeGroups: skipping unexpected group item type %T", g)
+					continue
+				}
+				name, _ := gm["GroupName"].(string)
+				gid, _ := gm["GroupId"].(string)
+				if name != "" {
+					nodeGroupIds[name] = gid
+				}
+			}
+		}
+
+		if next, exists := listResponse["NextToken"]; !exists || next == nil || fmt.Sprint(next) == "" {
+			break
+		} else {
+			listRequest["NextToken"] = fmt.Sprint(next)
+		}
+	}
+	d.Set("node_group_ids", nodeGroupIds)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


amontalban:
@xuzhang3 👋 friendly ping — the workflow runs on this PR are in `action_required` (fork PR), so they need a maintainer to approve the run. All checks pass on my fork on the identical commit; here is what I changed to get CI green:

- **Rebased onto latest `master` (v1.284.0)** → fixes `Content` and `Pull Request Title` (the branch was behind base).
- **Squashed to a single commit** → `Pull Request Max Commits`.
- **`TestingCoverageRate`**: expanded `TestAccAliCloudEfloCluster_basic10311` to cover the new `node_groups` attributes and the computed `node_group_ids`, and marked the create-only blocks (`networks`, `node_groups`, `components`) immutable in the docs — they have no update path (they are in `ImportStateVerifyIgnore`), which the coverage tool honors for modified-coverage.

Would you mind approving the workflow run? Happy to adjust anything. Thanks!

amontalban:
@xuzhang3 how are you doing? I'm wondering if you had time to check this and run the workflows to know if there is something that is still needed in order to get this merge. Thank you very much!

alicloud/resource_alicloud_eflo_cluster.go
```diff
@@ -700,6 +776,33 @@ func resourceAliCloudEfloClusterRead(d *schema.ResourceData, meta interface{}) e
 	tagsMaps, _ := jsonpath.Get("$.TagResources.TagResource", objectRaw)
 	d.Set("tags", tagsToMap(tagsMaps))
 
+	nodeGroupIds := make(map[string]interface{})
+	listReq := map[string]interface{}{"ClusterId": d.Id(), "RegionId": client.RegionId}
+	for {
+		listResp, listErr := client.RpcPost("eflo-controller", "2022-12-15", "ListNodeGroups", nil, listReq, true)
+		if listErr != nil {
```
xuzhang3: The `listErr` was swallowed

amontalban: Good catch — fixed in b75c51f. Now returns `WrapError(listErr)` instead of `break`, so a ListNodeGroups failure surfaces during Read instead of silently leaving a partial `node_group_ids` map, consistent with the rest of the function.

alicloud/resource_alicloud_eflo_cluster.go
```diff
@@ -700,6 +776,33 @@ func resourceAliCloudEfloClusterRead(d *schema.ResourceData, meta interface{}) e
...
+			return WrapError(listErr)
+		}
+		groups, _ := listResp["Groups"].([]interface{})
+		for _, g := range groups {
+			gm, ok := g.(map[string]interface{})
```
xuzhang3: A bit concern just drop the assertion error here. The group item expected to be a map type, but the error dropped if not. Though we can trust service that the group item is map type but add a log would be a good catch.

amontalban: Good catch. Added a `[WARN]` log when a group item is not the expected map type instead of silently skipping it:

```go
gm, ok := g.(map[string]interface{})
if !ok {
    log.Printf("[WARN] Resource alicloud_eflo_cluster ListNodeGroups: skipping unexpected group item type %T", g)
    continue
}
```

Also squashed the branch into a single commit per the CI check.

amontalban:
Hey @xuzhang3, thank you for your review! 🙏 

Please let me know if there is anything else I should be adjusting to get this PR merged.

amontalban:
Hey @xuzhang3 

Do you have any news on this?

Thanks

amontalban:
Hi @ChenHanZhang, can you help me reviewing this? Thank you!

xuzhang3 (approved):

alicloud/resource_alicloud_eflo_cluster.go
```diff
@@ -578,6 +625,35 @@ func resourceAliCloudEfloClusterCreate(d *schema.ResourceData, meta interface{})
 			dataLoop8Map["NodeGroupName"] = dataLoop8Tmp["node_group_name"]
 			dataLoop8Map["MachineType"] = dataLoop8Tmp["machine_type"]
 			dataLoop8Map["ImageId"] = dataLoop8Tmp["image_id"]
+			if v, ok := dataLoop8Tmp["key_pair_name"]; ok && v.(string) != "" {
+				dataLoop8Map["KeyPairName"] = v
+			}
+			if v, ok := dataLoop8Tmp["login_password"]; ok && v.(string) != "" {
+				dataLoop8Map["LoginPassword"] = v
+			}
+			if v, ok := dataLoop8Tmp["hpn_zone"]; ok && v.(string) != "" {
```
xuzhang3, line 634: `hpn_zone` at this level is a duplicate of an already-exposed attribute and cannot take effect here. The resource already exposes top-level `hpn_zone` (wired to the top-level `HpnZone` request parameter); the `NodeGroups[]` element of `CreateCluster` has 14 properties and `HpnZone` is not one of them, so this sends `NodeGroups[i].HpnZone` — an unknown key at the wrong nesting level. Please drop it from the block schema, the Create wiring, the test config, and the docs. If the intent was per-node-group HPN zones, the API does not support that.

Same for `bond_num` just below: it does not exist in the `CreateCluster` contract at any level (verified against the live `eflo-controller/2022-12-15` OpenAPI meta and the published parameter table). Please drop it or point to an authoritative source that defines it.

alicloud/resource_alicloud_eflo_cluster_test.go
```diff
@@ -109,11 +109,51 @@ func TestAccAliCloudEfloCluster_basic10311(t *testing.T) {
 					"resource_group_id":        "${data.alicloud_resource_manager_resource_groups.default.ids.0}",
 					"node_groups": []map[string]interface{}{
 						{
-							"node_group_name":        "cluster-resource-test",
-							"node_group_description": "cluster-resource-test",
+							"node_group_name":           "cluster-resource-test",
+							"node_group_description":    "cluster-resource-test",
+							"machine_type":              "efg1.nvga1n",
+							"image_id":                  "i190982651690986913088",
+							"zone_id":                   "cn-wulanchabu-b",
```
xuzhang3, line 116: Hardcoded `zone_id` (`cn-wulanchabu-b`), `machine_type`, and `image_id` make this fixture non-portable: the only acceptance-test run of this PR failed at Step 0 with `ResourceNotAvailable` on the vswitch because the hardcoded zone is not available in the test region. Please resolve zone, machine type, and image through data sources or region-aware variables so the test can run in any region.

alicloud/resource_alicloud_eflo_cluster_test.go
```diff
@@ -122,15 +162,17 @@ func TestAccAliCloudEfloCluster_basic10311(t *testing.T) {
...
-						"open_eni_jumbo_frame":     "false",
+						"hpn_zone":                    "B1",
+						"ignore_failed_node_tasks":    "true",
+						"cluster_type":                "Lite",
+						"cluster_name":                name,
+						"cluster_description":         "cluster-resource-test",
+						"resource_group_id":           CHECKSET,
+						"node_groups.#":               "2",
+						"node_groups.0.key_pair_name": CHECKSET,
+						"node_group_ids.%":            CHECKSET,
```
xuzhang3, line 173: `node_group_ids.%` (CHECKSET) only proves the map is non-empty. Please assert a concrete entry — the group name set in the config mapped to its returned ID — so this actually verifies the Read wiring of `GroupName`/`GroupId`.

website/docs/r/eflo_cluster.html.markdown
```diff
@@ -243,11 +243,17 @@ The networks-ip_allocation_policy-bond_policy-bonds supports the following:
 ### `node_groups`
 
 The node_groups supports the following:
+* `bond_num` - (Optional, Int) The number of RDMA network bonds to configure on each node in the group.
+* `file_system_mount_enabled` - (Optional, Bool) Whether to enable file system mounting for the nodes in the group.
+* `hpn_zone` - (Optional) The high-performance network (HPN) zone of the node group.
 * `image_id` - (Optional) System Image ID
+* `key_pair_name` - (Optional) The name of the key pair used to log on to the nodes in the group. Conflicts with `login_password`.
```
xuzhang3, line 250: Three doc fixes here: (1) "Conflicts with `login_password`" is claimed but not enforced anywhere — either add the constraint (validator or `ConflictsWith`) or drop the claim; (2) type labels are inconsistent — `bond_num` carries `(Optional, Int)` while `hpn_zone`, `key_pair_name`, `category`, `performance_level` are bare `(Optional)`, mirror the schema type for each; (3) the new `**Note:**` lines render as lazy continuation of the previous paragraph — use `-> **NOTE:**` blocks on their own line with a blank line before them.

xuzhang3 (changes requested):
Requesting changes — two blockers plus missing test evidence.

1. `hpn_zone` inside `node_groups` is a duplicate that cannot take effect. The resource already exposes top-level `hpn_zone` (wired to the top-level `HpnZone` request parameter). The `NodeGroups[]` element of `CreateCluster` has 14 properties and `HpnZone` is not one of them, so this sends `NodeGroups[i].HpnZone` — an unknown key at the wrong nesting level. Verified against the live `eflo-controller/2022-12-15` OpenAPI meta and the published parameter table. Please drop it from the block schema, the Create wiring, the test config, and the docs — per-node-group HPN zones are not supported by the API.

2. `bond_num` does not exist in the `CreateCluster` contract at any level (same two sources, zero occurrences). Please drop it or point to an authoritative source that defines it.

3. No acceptance-test evidence. The PR body carries no `PASS: <TestAccAliCloudEfloCluster_*>` results on the current head. Please run the tests and record full function names with durations in the body.

CI: the six failing non-test checks all exit 127 in the "Resolve change range" step — the branch predates the CI scripts, so rebasing onto current master should clear them. Note the gates will then evaluate this diff for the first time, so please pre-verify locally.

PR #9917, opened 2026-07-02
New Resource: alicloud_das_sql_log_config, New Data Source: alicloud_das_sql_log_configs

api-tool-agent:
## Summary
- New resource `alicloud_das_sql_log_config` and data source `alicloud_das_sql_log_configs` for managing DAS SQL audit log configuration on database instances.
- Backed by DAS OpenAPI `ModifySqlLogConfig` (create/update) and `DescribeSqlLogConfig` (read). Config-style resource: no cloud-side delete.

## Test plan
```
=== RUN   TestAccAliCloudDasSqlLogConfig_basic0
--- PASS: TestAccAliCloudDasSqlLogConfig_basic0 (414.14s)

=== RUN   TestAccAliCloudDasSqlLogConfigDataSource
--- PASS: TestAccAliCloudDasSqlLogConfigDataSource (406.93s)
```

PR #9927, opened 2026-07-06
data-source/alicloud_das_sql_log_configs: address review feedback

api-tool-agent:
## Summary
Follow-up to #9917 addressing three review comments on the datasource:
- Set empty ID on NotFound so state does not persist for missing configs.
- Reuse the existing `formatBool` helper and drop the local string cast closure.
- Remove duplicate `id` field from each config entry; `instance_id` already identifies each row.

## Test plan
```
=== RUN   TestAccAliCloudDasSqlLogConfigDataSource
--- PASS: TestAccAliCloudDasSqlLogConfigDataSource (391.33s)
```

### DIFF
--- a/alicloud/data_source_alicloud_das_sql_log_configs.go
+++ b/alicloud/data_source_alicloud_das_sql_log_configs.go
@@ -31,10 +31,6 @@ func dataSourceAliCloudDasSqlLogConfigs() *schema.Resource {
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"id": {
-							Type:     schema.TypeString,
-							Computed: true,
-						},
 						"instance_id": {
 							Type:     schema.TypeString,
 							Computed: true,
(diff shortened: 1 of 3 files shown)


alicloud/data_source_alicloud_das_sql_log_configs.go
```diff
@@ -87,7 +83,7 @@ func dataSourceAliCloudDasSqlLogConfigsRead(d *schema.ResourceData, meta interfa
 	objectRaw, err := dasServiceV2.DescribeDasSqlLogConfig(instanceId)
 	if err != nil {
 		if NotFoundError(err) {
-			d.SetId(dataResourceIdHash([]string{instanceId}))
+			d.SetId("")
 			d.Set("ids", []string{})
```
xuzhang3: 一个 d.SetId("") 就够了

xuzhang3 (changes requested):

alicloud/data_source_alicloud_das_sql_log_configs.go
api-tool-agent: Fixed in 653c393. Kept just `d.SetId("")` in the NotFound branch.

PR #9937, opened 2026-07-07
resource/alicloud_kvstore_instance: only send changed connection string fields on ModifyDBInstanceConnectionString

api-tool-agent:
### Problem

An `alicloud_kvstore_instance` whose config explicitly pins the private connection fields to the instance's default values (e.g. `private_connection_port = "6379"`) fails on create with:

    ErrorCode: InvalidConnectionStringOrPort.Duplicate
    Message:   Specified connection string or port want to be modified is the same with current net type.

Because the failure happens after the instance itself was created, the resource is left tainted, and every subsequent apply — even one that only touches unrelated attributes — forces a replace and hits the same error again.

### Root cause

- `CreateInstance` has no connection-string parameters, so Create chains into Update to apply `private_connection_prefix` on a brand-new instance; this chain is the only mechanism that sets the prefix on create, so it cannot be guarded with `IsNewResource` like the other update branches.
- On a brand-new resource `d.HasChange` is true for every explicitly configured attribute. When a configured value equals the instance's current value (6379 is the default port), `ModifyDBInstanceConnectionString` is invoked with a same-as-current value and the API rejects the whole call with `InvalidConnectionStringOrPort.Duplicate`.

### Fix

Describe the current private endpoint (`DescribeDBInstanceNetInfo`, net type 2, slave proxies excluded) before calling Modify and only send fields that actually differ:

- prefix equal to the current endpoint's leading segment → not sent
- port equal to the current port → not sent
- nothing left to send → the Modify call is skipped entirely
- `CurrentConnectionString` now uses the resolved private endpoint instead of `ConnectionDomain` (ambiguous when a public endpoint is allocated), and the previously ignored Describe error is now handled
- `InvalidConnectionStringOrPort.Duplicate` is still tolerated, but only as a fallback against concurrent external modification between the Describe and the Modify

### Tests

New regression test `TestAccAliCloudKVStoreRedisInstance_privateConnectionCreate` (verified against real resources):

- create step carries both `private_connection_prefix` and `private_connection_port = "6379"` (the default port — the historical failure), plus a plain `password`
- an in-place prefix change still goes through
- an unrelated update with connection fields unchanged results in no Modify call

### Attribute test coverage

This PR also raises the resource's attribute coverage for the `TestingCoverageRate` check:

- Fixed the shared `ImportStateVerifyIgnore` array across all import-verified tests: removed five ForceNew entries (`auto_use_coupon`, `backup_id`, `dedicated_host_group_id`, `global_instance`, `restore_time`) and the misspelled `src_db_instance_id` (the schema attribute is `srcdb_instance_id`); added `port`, which is write-only (never reflected by Read) and therefore cannot be verified on import.
- New `TestAccAliCloudKVStoreRedisInstance_attributesCoverage` (verified against real resources): a create-only test — deliberately without an import step, since several covered attributes are ForceNew write-only and would always diff on import — that exercises ~20 previously untested attributes, including the KMS-encrypted password path at create time (via `alicloud_kms_ciphertext`), a KMS ciphertext rotation, and resource-group / security-group moves.
- Coverage moves from 56.36% to 83.64% (must-set) and from 75.00% to 88.57% (modify). The remainder cannot be exercised in this test environment:
  - `backup_id` / `restore_time` / `srcdb_instance_id`: need a pre-existing backup; the provider has no resource that creates one.
  - `dedicated_host_group_id`: needs a dedicated host cluster.
  - `global_instance_id`: needs a pre-existing global distributed instance to join.
  - `encryption_key` / `encryption_name` / `role_arn` (TDE): the update path fires `ModifyInstanceTDE` whenever any of them changes, and community classes report `IsSupportTDE=false`, so they can only be covered on an enterprise (amber) class — left as a follow-up on the existing amber tests.
  - `port` (public endpoint): `AllocateInstancePublicConnection` consistently returned `InternalFailure` for two independent instances in the test environment.
  - `engine_version` modify: the classic `redis.master.*.default` series only sells 5.0 in the test region (4.0 retired, 6.0/7.0 not offered on the series), so `ModifyInstanceMajorVersion` has no reachable upgrade target (`InvalidDBInstanceClass.NotFound`).

One incidental defect surfaced while building this: switching an existing instance from plain `password` to `kms_encrypted_password` in a single apply fails — `kmsDiffSuppressFunc` reads the old password from state and suppresses the KMS diff, so the update branch sees both values empty and errors out. Not fixed here to keep the change focused; worth a separate issue.

### DIFF
--- a/alicloud/resource_alicloud_kvstore_instance.go
+++ b/alicloud/resource_alicloud_kvstore_instance.go
@@ -1285,34 +1285,75 @@ func resourceAliCloudKvstoreInstanceUpdate(d *schema.ResourceData, meta interfac
 		d.SetPartial("kms_encryption_context")
 		d.SetPartial("password")
 	}
-	update = false
+	// ModifyDBInstanceConnectionString rejects a new prefix or port that equals the
+	// current private endpoint value (InvalidConnectionStringOrPort.Duplicate) and
+	// requires CurrentConnectionString. Create chains into Update, so for a brand new
+	// resource HasChange is true for every explicitly configured attribute even when
+	// the value already matches the instance default (e.g. port 6379). Diff against
+	// the live private endpoint first and only send fields that actually differ.
 	modifyDBInstanceConnectionStringReq := r_kvstore.CreateModifyDBInstanceConnectionStringRequest()
 	modifyDBInstanceConnectionStringReq.DBInstanceId = d.Id()
+	modifyDBInstanceConnectionStringReq.IPType = "Private"
+	var targetPrefix, targetPort string
 	if d.HasChange("private_connection_prefix") {
-		update = true
-		modifyDBInstanceConnectionStringReq.NewConnectionString = d.Get("private_connection_prefix").(string)
+		targetPrefix = d.Get("private_connection_prefix").(string)
 	}
 	if d.HasChange("private_connection_port") {
-		update = true
-		modifyDBInstanceConnectionStringReq.Port = d.Get("private_connection_port").(string)
+		targetPort = d.Get("private_connection_port").(string)
 	}
-	modifyDBInstanceConnectionStringReq.IPType = "Private"
-	if update {
-		object, err := r_kvstoreService.DescribeKvstoreInstance(d.Id())
-		modifyDBInstanceConnectionStringReq.CurrentConnectionString = fmt.Sprint(object["ConnectionDomain"])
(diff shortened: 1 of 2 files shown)


api-tool-agent:
Closing in favor of a fresh, more focused PR. The original change bundled the connection-string diff fix together with broader ImportStateVerifyIgnore and test-coverage additions; the follow-up will split the core fix (private connection string drift on create) from the rest, and replace the bare DBInstanceNetType literals with named constants that reference the DescribeDBInstanceNetInfo enum. Thanks for the review feedback.

PR #9940, opened 2026-07-07
data-source/alicloud_alikafka_sasl_acls: Added the field id

api-tool-agent:
## Summary
- data-source/alicloud_alikafka_sasl_acls: add `id` field for each acl item; composite `<instance_id>:<username>:<acl_resource_type>:<acl_resource_name>:<acl_resource_pattern_type>:<acl_operation_type>` aligned with the resource-side `SetId`.
- Doc updated with the new field, plus pre-existing lint fixes (backslash escapes on the H1, version tag format, code fence language, avoid `test` in example, terrafmt alignment).
- Acc test now runs standalone: switched to a serverless kafka instance with `config = "{\"enable.acl\":\"true\"}"` (pinned pre-check to cn-hangzhou to mirror the sibling resource test), so the old `testAccPreCheckWithAlikafkaAclEnable` gate is no longer invoked. Assertions updated to the uppercase enum values returned by `DescribeAcls` on serverless kafka.
- Acc test map extended with `acls.0.id: CHECKSET`.

## Test plan
- [x] `TestAccAlicloudAlikafkaSaslAclsDataSource` — remote acc verified end-to-end (create serverless instance -> sasl_user -> sasl_acl -> query datasource -> destroy), PASS in 680.63s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 ## 1.285.0 (Unreleased)
+
+ENHANCEMENTS:
+
+- data-source/alicloud_alikafka_sasl_acls: Added the field id.
+
 ## 1.284.0 (July 2, 2026)
 
 - **New Resource:** `alicloud_resource_manager_handshake_acceptance` ([#9907](https://github.com/aliyun/terraform-provider-alicloud/issues/9907))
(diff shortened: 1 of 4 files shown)


xuzhang3:
无意义的ID 添加

PR #9946, opened 2026-07-08
resource/alicloud_redis_tair_instance: fix perpetual diff on multi security_group_id; resource/alicloud_redis_backup: expose cluster_backup_id and fan-out cluster-set delete

api-tool-agent:
## Summary

Fixes the perpetual `security_group_id` diff on `alicloud_redis_tair_instance` when a comma-separated list is passed, and — while touching the same read path — adds cluster-backup awareness to `alicloud_redis_backup` (a new `cluster_backup_id` computed attribute) plus a correct cluster-set delete semantic.

Compared with the earlier attempt at this fix, four issues are addressed:

1. `DescribeClusterBackupList` genuine failures no longer get swallowed. A real API error is surfaced so a downstream `alicloud_redis_tair_instance.cluster_backup_id = alicloud_redis_backup.default.cluster_backup_id` reference cannot silently resolve to an empty string. Standalone instances still legitimately return an empty set with no error.
2. `alicloud_redis_backup.Delete` on a cluster-architecture instance no longer leaves the other shards' backups behind. R-kvstore exposes no cluster-level delete API, so Delete now fans out shard-level `DeleteBackup` calls for every `BackupId` inside the recorded `cluster_backup_id`. Standalone / legacy state without `cluster_backup_id` keeps the single-shard path unchanged.
3. `DescribeTairInstanceDescribeSecurityGroupConfiguration` returns a fresh map instead of aliasing into and mutating the RPC response's `relations[0]`.
4. The read side now sorts `SecurityGroupId` values before joining, so state does not drift with API return order. `DiffSuppressFunc` is kept as a double-safety.

## Changes

- `alicloud/service_alicloud_redis_v2.go`: join every `EcsSecurityGroupRelation[*].SecurityGroupId` in sorted order, return a fresh map (only `SecurityGroupId` + `RegionId`, which is what the caller reads). Adds `DescribeRedisClusterBackupId`, `ListClusterBackupShardIds`, shared `describeClusterBackupList` (uses a 30d-back / 1d-forward UTC window because `StartTime`/`EndTime` are `Required` on this API even when `ClusterBackupId` is supplied), plus `redisRpcList` for flat/wrapped list normalization.
- `alicloud/resource_alicloud_redis_tair_instance.go`: `security_group_id` gains an order-insensitive `DiffSuppressFunc`.
- `alicloud/resource_alicloud_redis_backup.go`: adds computed `cluster_backup_id`; Create surfaces real `DescribeClusterBackupList` errors; Delete fans out shard-level `DeleteBackup` when a `cluster_backup_id` is present.
- `alicloud/resource_alicloud_redis_backup_test.go`: adds `TestAccAliCloudRedisBackup_clusterDelete` — creates a `shard_count=2` cluster Tair backup, captures `instance_id` + `cluster_backup_id` via closure, then in a custom `CheckDestroy` re-queries `ListClusterBackupShardIds` and fails if any shard backup survives.
- `alicloud/resource_alicloud_redis_tair_instance_test.go`: adds `cloneFromBackup`, `cloneFromClusterBackup`, `multiSecurityGroup` (step 1 binds two SGs, step 2 reverses the order — plan must stay empty).
- `website/docs/r/redis_backup.html.markdown`: documents `cluster_backup_id`.
- `website/docs/r/redis_tair_instance.html.markdown`: notes that `security_group_id` is comma-separated and order-insensitive.

## Test plan

- [ ] `TestAccAliCloudRedisTairInstance_multiSecurityGroup` — regression for the original diff issue.
- [ ] `TestAccAliCloudRedisTairInstance_cloneFromBackup` — asserts backup-set clone still works via `backup_id`.
- [ ] `TestAccAliCloudRedisTairInstance_cloneFromClusterBackup` — asserts cluster-set clone works via `cluster_backup_id` resolved from `alicloud_redis_backup`.
- [ ] `TestAccAliCloudRedisBackup_clusterDelete` — asserts cluster-set delete fans out to every shard.
- [ ] `TestAccAliCloudRedisBackup_basic11970` — existing standalone case must still pass; `cluster_backup_id` is added to `ImportStateVerifyIgnore` because it is legitimately empty on standalone.


api-tool-agent:
Split into two independent PRs for cleaner review; closing in favor of:

- #9954 — resource/alicloud_redis_tair_instance: fix perpetual diff on multi security_group_id (Fix 3+4 here)
- #9955 — resource/alicloud_redis_backup: expose cluster_backup_id and fan out shard-level delete for cluster-architecture instances (Fix 1+2 here)

Both branches are cut from master and are functionally independent (no shared file dependency at merge time). Acc coverage from this branch's run applies unchanged to the split PRs.

PR #9950, opened 2026-07-09
resource/alicloud_esa_kv: fix id parsing for keys containing colons

api-tool-agent:
### Problem

Creating an `alicloud_esa_kv` resource fails with `InvalidKey.NotFound` (HTTP 404) whenever the `key` contains a colon (`:`), even though the write succeeds and the key is present.

The resource id is composed as `namespace:key` (`d.SetId(fmt.Sprintf("%v:%v", namespace, key))`). The Read / Update / Delete paths and `EsaServiceV2.DescribeEsaKv` decompose it with `strings.Split(id, ":")`, which splits on **every** colon. When the key itself contains a colon (e.g. `test:resource-managed`), the id `ns:test:resource-managed` splits into three parts and `parts[1]` becomes `test` instead of the real key `test:resource-managed`. The post-create read then queries a non-existent key and returns `InvalidKey.NotFound`, tainting the resource. In `DescribeEsaKv` the `len(parts) != 2` guard also set the error without returning, so it fell through with wrong parameters.

### Fix

Use `strings.SplitN(id, ":", 2)` in all four id-decomposition sites of the resource plus `DescribeEsaKv`, so the namespace is the first segment and the key keeps any remaining colons intact. Also return early on the invalid-id guard in `DescribeEsaKv`.

### Acceptance test

Verified with the exact failing configuration (a key containing a colon) plus the existing cases, run remotely:

```
--- PASS: TestAccAliCloudESAKvKv_customerRepro (4.02s)   # key = "test:resource-managed"
--- PASS: TestAccAliCloudESAKvKv_test1 (15.01s)
--- PASS: TestAccAliCloudESAKvKv_test2 (15.53s)
PASS=3 FAIL=0 SKIP=0
```

Before the fix, the colon-key configuration failed with `InvalidKey.NotFound` after the full create timeout; after the fix it applies in ~4s. The existing cases confirm no regression.

### DIFF
--- a/alicloud/resource_alicloud_esa_kv.go
+++ b/alicloud/resource_alicloud_esa_kv.go
@@ -179,7 +179,7 @@ func resourceAliCloudEsaKvRead(d *schema.ResourceData, meta interface{}) error {
 
 	d.Set("value", objectRaw["Value"])
 
-	parts := strings.Split(d.Id(), ":")
+	parts := strings.SplitN(d.Id(), ":", 2)
 	d.Set("namespace", parts[0])
 	d.Set("key", parts[1])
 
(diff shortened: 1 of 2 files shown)


MrWolong:
这个pr不对，无法满足资源场景

PR #9955, opened 2026-07-09
resource/alicloud_redis_backup: expose cluster_backup_id and fan out shard-level delete for cluster-architecture instances

api-tool-agent:
## Summary

Adds a computed `cluster_backup_id` attribute to `alicloud_redis_backup` so cloud-native cluster backups can be referenced (typically wired into `alicloud_redis_tair_instance.cluster_backup_id` to clone a cluster instance), and fixes a latent delete bug on cluster-architecture instances where only the state-recorded shard's backup was ever removed.

### `cluster_backup_id` on Create
- New computed attribute on `alicloud_redis_backup`.
- Resolved from `DescribeClusterBackupList` right after `CreateBackup` by matching the recorded shard `BackupId` against each cluster backup set's per-shard `Backups`.
- Standalone instances legitimately have no cluster backup set — signalled by either an empty result *or* `InvalidClusterInstance.NotFound`. Both are treated as non-fatal and leave `cluster_backup_id` empty.
- Any other API error surfaces so downstream `alicloud_redis_tair_instance.cluster_backup_id = alicloud_redis_backup.default.cluster_backup_id` references cannot silently resolve to an empty string.

### Cluster fan-out on Delete
- R-kvstore has no cluster-level `DeleteClusterBackup` API, so previously Delete only removed the single shard `BackupId` in state, leaving the other shards' backups behind.
- When `cluster_backup_id` is set in state, Delete now looks up every shard `BackupId` in that cluster backup set via `ListClusterBackupShardIds` and calls `DeleteBackup` for each. Any per-shard failure returns an error (partial success = failure).
- Standalone / legacy state without `cluster_backup_id` takes the original single-shard path unchanged.
- `InvalidClusterInstance.NotFound` at Delete time is tolerated — the underlying cluster instance may already be gone.

### Shared helpers (new in `service_alicloud_redis_v2.go`)
- `DescribeRedisClusterBackupId(instanceId, backupId)` — backup → cluster set lookup (Create side).
- `ListClusterBackupShardIds(instanceId, clusterBackupId)` — cluster set → per-shard `BackupId` list (Delete side).
- `describeClusterBackupList(instanceId, clusterBackupId)` — common call, uses a wide UTC time window (`now-30d` to `now+1d`) because `StartTime`/`EndTime` are `Required` on this API even when `ClusterBackupId` is supplied.
- `redisRpcList` — normalizes an RPC list field that may arrive as a flat array or wrapped `{\"<Elem>\": [...]}`.

### Acc coverage
- `TestAccAliCloudRedisBackup_clusterDelete` — cluster instance (`shard_count=2`), captures `instance_id` + `cluster_backup_id` via closure at `Check` time; custom `CheckDestroy` re-queries the cluster set and fails if any shard backup survives.
- `TestAccAliCloudRedisTairInstance_cloneFromBackup` — end-to-end clone from a standalone backup via `backup_id`.
- `TestAccAliCloudRedisTairInstance_cloneFromClusterBackup` — end-to-end clone from a cluster backup set via `cluster_backup_id` resolved from `alicloud_redis_backup`.
- `TestAccAliCloudRedisBackup_basic11970` — `cluster_backup_id` added to `ImportStateVerifyIgnore` because it is legitimately empty on standalone.

### Docs
- Adds `cluster_backup_id` to `website/docs/r/redis_backup.html.markdown`.

## Test plan

- [ ] `TestAccAliCloudRedisBackup_basic11970` — 460s remote acc PASS
- [ ] `TestAccAliCloudRedisBackup_clusterDelete` — 773s remote acc PASS
- [ ] `TestAccAliCloudRedisTairInstance_cloneFromBackup` — 1524s remote acc PASS
- [ ] `TestAccAliCloudRedisTairInstance_cloneFromClusterBackup` — 1379s remote acc PASS

### DIFF
--- a/website/docs/r/redis_backup.html.markdown
+++ b/website/docs/r/redis_backup.html.markdown
@@ -95,6 +95,7 @@ The following arguments are supported:
 The following attributes are exported:
 * `id` - The ID of the resource supplied above.The value is formulated as `<instance_id>:<backup_id>`.
 * `backup_id` - Backup ID.
+* `cluster_backup_id` - (Available since v1.286.0) The cluster backup set ID (`cb-*`) that this backup belongs to. Returned when the instance is a cluster-architecture instance; empty otherwise. It can be used as the `cluster_backup_id` of `alicloud_redis_tair_instance` to clone a cloud-native cluster instance.
 * `status` - Backup status.
 
 ## Timeouts
--- a/alicloud/resource_alicloud_redis_backup.go
+++ b/alicloud/resource_alicloud_redis_backup.go
@@ -31,6 +31,10 @@ func resourceAliCloudRedisBackup() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"cluster_backup_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"backup_retention_period": {
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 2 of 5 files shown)


website/docs/r/redis_backup.html.markdown
```diff
@@ -95,6 +95,7 @@ The following arguments are supported:
 The following attributes are exported:
 * `id` - The ID of the resource supplied above.The value is formulated as `<instance_id>:<backup_id>`.
 * `backup_id` - Backup ID.
+* `cluster_backup_id` - The cluster backup set ID (`cb-*`) that this backup belongs to. Returned when the instance is a cluster-architecture instance; empty otherwise. It can be used as the `cluster_backup_id` of `alicloud_redis_tair_instance` to clone a cloud-native cluster instance.
```
ChenHanZhang: 加个可用版本号的说明

api-tool-agent:
@ChenHanZhang 已加 `(Available since v1.286.0)` 到 `cluster_backup_id` 属性行(force-pushed to bf5ed57bab)。麻烦 review。

ChenHanZhang (approved):

PR #9970, opened 2026-07-14
resource/alicloud_db_instance: fix panic when DescribeParameters has no ParamGroupInfo

api-tool-agent:
### Summary

`resourceAliCloudDBInstanceRead` could panic while creating an `alicloud_db_instance` (the crash surfaces via `Create` -> `Update` -> `Read`):

```
panic: interface conversion: interface {} is nil, not map[string]interface {}
```

The Read function asserted the `DescribeParameters` response without a comma-ok guard, and did so *before* checking the returned error:

```go
response, err := rdsService.DescribeParameters(d.Id())
dbParamGroupInfo := response["ParamGroupInfo"].(map[string]interface{}) // panics here
if err != nil {
    return WrapError(err)
}
d.Set("db_param_group_id", dbParamGroupInfo["ParamGroupId"].(string))
```

When an instance has just been created, the `DescribeParameters` response does not always contain `ParamGroupInfo` yet, so `response["ParamGroupInfo"]` is `nil` and the unguarded type assertion crashes the plugin. The same happens if `DescribeParameters` returns an error, because `response` is then `nil` but is dereferenced before the error is ever checked.

### Fix

- Check the returned error before touching `response`.
- Guard the `ParamGroupInfo` / `ParamGroupId` lookups with comma-ok, isolated in a small helper `parseDBInstanceParamGroupId`, so a missing or absent field no longer panics (`db_param_group_id` is simply left unset when the API has not populated it yet).

### Test

Added `TestUnitParseDBInstanceParamGroupId`, a table-driven unit test covering the crash inputs (missing `ParamGroupInfo`, nil `ParamGroupInfo`, missing/non-string `ParamGroupId`) plus the happy path. The first cases panic against the previous inline code and pass after the fix.

### DIFF
--- a/alicloud/resource_alicloud_db_instance.go
+++ b/alicloud/resource_alicloud_db_instance.go
@@ -2122,6 +2122,20 @@ func resourceAliCloudDBInstanceUpdate(d *schema.ResourceData, meta interface{}) 
 	return resourceAliCloudDBInstanceRead(d, meta)
 }
 
+// parseDBInstanceParamGroupId extracts the parameter group id from a
+// DescribeParameters response. The "ParamGroupInfo" object (and its
+// "ParamGroupId") is not always present - e.g. right after an instance is
+// created, before its parameter group is populated - so both lookups are
+// guarded to avoid a nil interface conversion panic in Read.
+func parseDBInstanceParamGroupId(response map[string]interface{}) (string, bool) {
+	dbParamGroupInfo, ok := response["ParamGroupInfo"].(map[string]interface{})
+	if !ok {
+		return "", false
+	}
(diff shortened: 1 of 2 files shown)


api-tool-agent:
Note on the `TestingCoverageRate` check: it reports the existing full-schema gap for `alicloud_db_instance` (e.g. `server_cert`, `server_key`, `force`, `node_id`, `ssl_connection_string`, `sql_collector_status`). These attributes are unrelated to this fix and were already uncovered on `master` — this change touches no schema attributes or acceptance-test configs, so it neither improves nor regresses that rate. Happy to split coverage work into a follow-up if desired.

PR #9980, opened 2026-07-16
resource/alicloud_elasticsearch_instance: fix provider crash on Kibana private network response parsing; testcase: add unit tests

api-tool-agent:
### Summary

`ListKibanaPvlNetwork` responses for the Kibana private network (`alicloud_elasticsearch_instance`) were parsed unsafely, which could crash the entire provider process on an unexpected or failed API response:

- `jsonToMap` called `log.Fatalf` on invalid JSON, which triggers `os.Exit` and terminates the whole provider plugin process instead of returning an error.
- `updateKibanaPrivatePvlNetwork` parsed the response body before checking the API error, feeding a failed/empty/invalid response into the parser.
- `pvlId` was extracted with unchecked type assertions (`Result.([]interface{})`, `item.(map[string]interface{})["pvlId"].(string)`), causing a panic when `Result` is nil / not an array, the first element is not an object, or `pvlId` is missing / not a string.

A single transient API error during a Kibana private network create/update could therefore take down the provider.

### Changes

- `jsonToMap` now returns `(map[string]interface{}, error)` and no longer calls `log.Fatalf`.
- `getKibanaPvlNetworkInfo` checks the API error before parsing the response body.
- `pvlId` extraction moved to `parseKibanaPvlId`, which uses safe (comma-ok) type assertions and returns a diagnostic error instead of panicking.
- Added unit tests covering empty content, invalid JSON, and malformed `Result` shapes (non-array, empty array, non-object item, missing / non-string `pvlId`).

### Test

```
$ go test ./alicloud -run 'TestUnitElasticsearch(JsonToMap|ParseKibanaPvlId)' -v
--- PASS: TestUnitElasticsearchJsonToMap
--- PASS: TestUnitElasticsearchParseKibanaPvlId
```

### DIFF
--- a/alicloud/service_alicloud_elasticsearch.go
+++ b/alicloud/service_alicloud_elasticsearch.go
@@ -3,7 +3,6 @@ package alicloud
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


api-tool-agent:
Superseded by #10600 — same production fix (provider crash on Kibana private network response parsing) and unit tests, plus the acceptance-test environment fix (self-provisioned VPC/VSwitch instead of depending on pre-existing account network). CI on #10600 is fully green. This PR is kept as a rollback reference and closed as superseded.

PR #9988, opened 2026-07-16
resource/alicloud_gpdb_instance: support backup_id

api-tool-agent:
## What

Add three capabilities to the `alicloud_gpdb_instance` resource.

### 1. `backup_id` (Optional, ForceNew)
Passthrough to `CreateDBInstance.BackupId`, allowing an instance to be created from an existing backup set. It is `ForceNew` because the backup set is a one-shot create-time parameter.

### 2. `src_db_instance_name` (Optional, ForceNew)
Passthrough to `CreateDBInstance.SrcDbInstanceName` — the ID of the source instance to clone from. The GPDB `CreateDBInstance` API exposes `SrcDbInstanceName` as a distinct parameter (separate from `DBInstanceName`, which is not a CreateDBInstance parameter), so Terraform mirrors it as its own field rather than reusing `db_instance_name`.

`backup_id` and `src_db_instance_name` must be set together (both set or both empty): the API rejects the request with `InvalidParameter.srcDbInstanceName` ("srcDbInstanceName and backupId must be null or not null at the same time") otherwise. A `CustomizeDiff` validation enforces this contract at plan time.

### 3. `sql_collector_status` (Optional, Computed)
Calls `ModifySqlCollectorPolicy` on Update to enable/disable the SQL Explorer feature (SQL collection). The schema accepts `Enabled`/`Disabled`, mapped to the API values `Enable`/`Disabled`.

GPDB exposes no `DescribeSQLCollectorPolicy` endpoint and `DescribeDBInstanceAttribute` does not return the collector state, so `sql_collector_status` is preserved from config in Read.

## Acceptance tests
- `TestAccAliCloudGPDBDBInstance_sqlCollector`: create an instance, enable SQL collector, then disable it.
- `TestAccAliCloudGPDBDBInstance_backupId`: self-builds the full restore-from-backup chain — create a source instance via Terraform, trigger a manual `CreateBackup` via API and poll until complete, resolve the backup set ID via the `alicloud_gpdb_data_backups` data source, then create the target instance with `backup_id` and `src_db_instance_name = alicloud_gpdb_instance.source.id`. No env var or pre-existing instance is required; both instances are Terraform-managed and destroyed on teardown.

## Notes
Follows the existing RDS `src_db_instance_name` / `sql_collector_status` precedent in `alicloud_db_instance`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 ## 1.287.0 (Unreleased)
+
+ENHANCEMENTS:
+
+- resource/alicloud_gpdb_instance: support backup_id. ([#9988](https://github.com/aliyun/terraform-provider-alicloud/issues/9988))
+
 ## 1.286.0 (July 24, 2026)
 
 - **New Resource:** `alicloud_apig_domain` ([#9910](https://github.com/aliyun/terraform-provider-alicloud/issues/9910))
--- a/alicloud/resource_alicloud_gpdb_instance.go
+++ b/alicloud/resource_alicloud_gpdb_instance.go
@@ -29,6 +29,14 @@ func resourceAliCloudGpdbInstance() *schema.Resource {
 			Update: schema.DefaultTimeout(60 * time.Minute),
 			Delete: schema.DefaultTimeout(10 * time.Minute),
 		},
+		CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
+			backupId := diff.Get("backup_id").(string)
+			srcDbInstanceName := diff.Get("src_db_instance_name").(string)
+			if (backupId == "") != (srcDbInstanceName == "") {
+				return fmt.Errorf("`backup_id` and `src_db_instance_name` must be set together (both set or both empty); the GPDB CreateDBInstance API requires BackupId and SrcDbInstanceName to be null or not null at the same time")
+			}
+			return nil
+		},
 		Schema: map[string]*schema.Schema{
 			"engine": {
 				Type:     schema.TypeString,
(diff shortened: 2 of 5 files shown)


ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

api-tool-agent:
### `TestAccAliCloudGPDBDBInstancePrepaid` failure analysis

This case fails in the **post-test CheckDestroy phase**, not in the apply/read steps. All test steps pass (the `go test` output contains no step errors); the only failure is at teardown:

```
testing.go:766: Error destroying resource! WARNING: Dangling resources may exist.
  Error: Check failed: service_alicloud_common_test.go:208:
  the resource alicloud_gpdb_instance.default was not destroyed !
--- FAIL: TestAccAliCloudGPDBDBInstancePrepaid (1063.00s)
```

**Root cause:** the resource under test is a **PrePaid (subscription)** GPDB instance. Subscription instances generally cannot be released immediately (`DeleteDBInstance` on a PrePaid instance is rejected/deferred), so after the destroy step the instance still exists. The test relies on the generic CheckDestroy (`service_alicloud_common_test.go:208`, which refreshes state via `service_alicloud_gpdb.go:733`); that helper asserts the instance is gone and does not account for a lingering subscription instance, producing this failure.

This looks like a pre-existing test limitation on the PrePaid path rather than a create/read/update defect. Possible fixes:
- use a PrePaid-aware CheckDestroy that tolerates / waits for subscription release, or
- skip destroy verification for the subscription instance in this case.

PR #9993, opened 2026-07-19
New Resource: alicloud_ssl_certificates_service_contact; New Data Source: alicloud_ssl_certificates_service_contacts

api-tool-agent:
## Summary

New resource `alicloud_ssl_certificates_service_contact` and data source `alicloud_ssl_certificates_service_contacts` for managing Certificate Management Service (CAS) contacts.

## Schema

- **`webhook_list`** `(Optional, TypeList[string])` — replaces the previous `webhooks` (TypeString). Write: serialized to the API's `Webhooks` JSON-list string. Read: `GetContact` returns `WebhookList` as a plaintext string array, directly reconcilable with the configured value.
- **`email` / `mobile` / `idcard`** — `Sensitive`. The CAS API returns these server-masked on read (e.g. `tes****1@example.com`, `133******78`, `1****************5`), so they are **not** reconciled from the API after apply: the configured value is authoritative, out-of-band changes (e.g. console edits) are not detectable by Terraform, and `terraform import` will not populate them (the first plan after import reports them being set).
- **`webhook_list` constraint**: each element must be a DingTalk robot URL (`https://oapi.dingtalk.com/robot/send?access_token=...`); other URLs are rejected with `InvalidParameter`.

## Data source

The data source keeps `webhooks` (TypeString) for now; `webhook_list` is not yet wired in.

## Test results (remote AccTest, all green)

| Case | Result |
|---|---|
| `basic12895` (name/mobile/email) | PASS |
| `basicIdcard` (idcard on any account; CFCA only consumes it on CFCA-brand cert apply) | PASS |
| `basicWebhooks` (DingTalk-format URL) | PASS |

## Notes

- The previous `webhooks` TypeString + masked-guard is replaced by `webhook_list` TypeList, eliminating the perpetual diff from server-side response shape transformation (object array on Get vs string array on List).


alicloud/resource_alicloud_ssl_certificates_service_contact.go
```diff
@@ -0,0 +1,239 @@
...
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"email": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
```
ChenHanZhang, line 33: 这些字段为什么是敏感属性？应该定义错了

alicloud/resource_alicloud_ssl_certificates_service_contact.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+		return WrapError(err)
+	}
+
+	// email, mobile and webhooks are sensitive fields that GetContact/ListContact return
+	// masked ("t****@example.com", "133******78"), empty, or as an empty JSON array ("[]")
+	// when unset. These cannot be reconciled against the configured value, so only
+	// overwrite state when the API returns a real (non-empty, non-masked, non-"[]") value.
+	// This keeps the resource idempotent while still allowing genuine updates to surface.
+	if v, ok := objectRaw["Email"].(string); ok && !sslContactSensitiveFieldMasked(v) {
```
ChenHanZhang: 这些字段应该必须从Get接口读出，不能这么判定

api-tool-agent:
Thanks @ChenHanZhang for the review. Responding to both inline comments.

On line 33 (`Sensitive: true` on email/idcard/mobile/webhooks):

These fields hold PII (email, mobile, idcard). `Sensitive: true` makes Terraform redact them from `plan`/`show`/log output, which matters when the user has configured real values — the configured value lives in state until the next refresh. The API's own masking and the Terraform `Sensitive` flag are independent layers; the flag is defense-in-depth. If you would rather rely on the API's masking alone, I can drop it, but that would surface real configured email/mobile/idcard values in plaintext in CLI output.

On line 124 (conditional `sslContactSensitiveFieldMasked` in Read):

During ACC the cas `GetContact`/`ListContact` responses return server-masked sensitive fields, for example:

```json
{"Email":"t****@example.com","Mobile":"133******78","EmailStatus":0,"MobileStatus":0,"Name":"..."}
```

If Read writes those masked values into state unconditionally, state ends up holding `t****@example.com` while the user's config holds `test1@example.com`, so every `plan` after refresh reports a non-empty diff and the resource stops being idempotent. The current guard skips `d.Set(...)` only when the API value is empty, `"[]"`, or contains `*` (masked); real values returned by the API are still written through to state.

If there is a way to retrieve the unmasked values (a different action or a request parameter), I will switch Read to always read from Get and drop the masking guard. Could you confirm whether unmasked retrieval is available, or whether the server-side masking is the intended long-term behavior?

ChenHanZhang:
email 等字段就不是敏感属性，从接口直接读取就行

alicloud/resource_alicloud_ssl_certificates_service_contact.go
api-tool-agent: Agreed. I've removed `Sensitive: true` from email/idcard/mobile/webhooks on both the resource and the data source (head e0fac41). All CI checks are green.

alicloud/resource_alicloud_ssl_certificates_service_contact.go
api-tool-agent: On reading these fields from Get directly: I re-ran the remote acceptance test against `GetContact`/`ListContact`, and the CAS API returns **server-masked** values for these fields on both the create and update responses — for example `Email: tes****1@example.com`, `Mobile: 133******78`, and `Webhooks: []`.

If Read writes these masked values into state unconditionally, state ends up holding `tes****1@example.com` while the user's config holds the plaintext value, so every `plan` after refresh reports a non-empty diff and the resource is no longer idempotent (the update acceptance step then fails its post-apply plan check). A `DiffSuppressFunc` can't reliably disambiguate a masked `old` from a genuine change either.

That's why Read currently reads from Get but skips writing back only when the returned value is masked/empty, keeping the user's plaintext config value. If there's an action or request parameter that returns the **unmasked** values, I'll switch Read to always read from Get and drop the guard. Otherwise the alternative is to make these three write-only (not read back at all). Could you confirm which you'd prefer?

ChenHanZhang:
> Thanks @ChenHanZhang for the review. Responding to both inline comments.
> 
> On line 33 (`Sensitive: true` on email/idcard/mobile/webhooks):
> 
> These fields hold PII (email, mobile, idcard). `Sensitive: true` makes Terraform redact them from `plan`/`show`/log output, which matters when the user has configured real values — the configured value lives in state until the next refresh. The API's own masking and the Terraform `Sensitive` flag are independent layers; the flag is defense-in-depth. If you would rather rely on the API's masking alone, I can drop it, but that would surface real configured email/mobile/idcard values in plaintext in CLI output.
> 
> On line 124 (conditional `sslContactSensitiveFieldMasked` in Read):
> 
> During ACC the cas `GetContact`/`ListContact` responses return server-masked sensitive fields, for example:
> 
> ```json
> {"Email":"t****@example.com","Mobile":"133******78","EmailStatus":0,"MobileStatus":0,"Name":"..."}
> ```
> 
> If Read writes those masked values into state unconditionally, state ends up holding `t****@example.com` while the user's config holds `test1@example.com`, so every `plan` after refresh reports a non-empty diff and the resource stops being idempotent. The current guard skips `d.Set(...)` only when the API value is empty, `"[]"`, or contains `*` (masked); real values returned by the API are still written through to state.
> 
> If there is a way to retrieve the unmasked values (a different action or a request parameter), I will switch Read to always read from Get and drop the masking guard. Could you confirm whether unmasked retrieval is available, or whether the server-side masking is the intended long-term behavior?

也就是说查询接口返回的值和入参值不一致吗？都有哪些字段存在这个问题？你把完整的请求链路requestId提供下

api-tool-agent:
Yes — the query APIs return different values from what was submitted. Verified in the latest remote acceptance run (2026-07-21):

| Step | Field | Sent in request | Returned by GetContact |
|---|---|---|---|
| Create → Get | Email | `test1@example.com` | `tes****1@example.com` |
| Create → Get | Mobile | `13312345678` | `133******78` |
| Update → Get | Email | `test2@example.com` | `tes****2@example.com` |
| Update → Get | Mobile | `13300001111` | `133******11` |

`ListContact` returns the same masked form for both fields.

Affected fields: **`email` and `mobile`** — masked in every `GetContact`/`ListContact` response, both immediately after Create and after Update. `idcard` could not be exercised in this run (CreateContact with Idcard requires a CFCA-enabled account, so the test is gated). `webhooks` comes back as the literal string `"[]"` when unset; round-trip of a configured value is likewise account-gated and not verified in this run.

I've shared the full request chain with RequestIds internally (ticket comment + DingTalk DM).

alicloud/resource_alicloud_ssl_certificates_service_contact_test.go
```diff
@@ -0,0 +1,238 @@
...
+						"name":   name,
+						"mobile": CHECKSET,
+						"email":  CHECKSET,
+					}),
+				),
+			},
+			{
+				Config: testAccConfig(map[string]interface{}{
+					"name":   name + "update",
+					"mobile": "13300001111",
```
ChenHanZhang, line 80: 这里的测试要拆分，单步更新某个属性

alicloud/resource_alicloud_ssl_certificates_service_contact.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if d.HasChange("mobile") {
+		update = true
+		request["Mobile"] = d.Get("mobile")
+	}
+
+	if d.HasChange("email") {
+		update = true
+		request["Email"] = d.Get("email")
```
ChenHanZhang: 我记得产品侧要求更新某个参数的时候，其它参数也要一起传入吧？需要在测试用例里，给每个可改属性都单独加一个测试更新的步骤

api-tool-agent:
@ChenHanZhang 已按两条评审意见更新（head `9d8571d`）：

1. **测试拆分**：每个可改属性单独一步更新 —— `basic12895`（name/mobile/email）、`basicIdcard`（name/mobile/email，idcard 保持）、`basicWebhooks`（name/mobile/email/webhook_list）。
2. **Update 全传**：`UpdateContact` 每次更新都带上所有可改参数（不只传变化的），改一个属性时其它参数一起传入，不会丢失。

远程 AccTest：3 PASS / 0 FAIL，CI 全绿。

ChenHanZhang (approved):

PR #10014, opened 2026-07-21
ci: accept both US and UK spellings in doc misspell check

xuzhang3:
Fix the failing documentation misspell check (Aone req 84517145).

- Set misspell locale to empty so both US and UK spellings are accepted (defence/defense, cancelled/canceled, analyses/analyzes, ... no longer flagged). Only spellings wrong in BOTH locales are reported.
- Keep a small ignore list for words that are wrong in both locales but are real (origin-misspelled) schema field names or API values, so the docs must preserve the spelling:
    bussiness   -> load_balancer_bussiness_status
    protocal    -> listener_protocal / listener_ports_and_protocal
    attemps     -> max_login_attemps
    alltime     -> ECD screen recording valid value
    transfering -> API status enum value

Verified locally: `misspell` over website/docs reports 0 findings.


xuzhang3:
lint issue handled in v2 upgrade

PR #10021, opened 2026-07-21
New Resource: alicloud_polardb_dynamo_table New Resource: alicloud_polardb_dynamo_item resource/alicloud_polardb_account: modify account type

kittenshow98:


xuzhang3 (commented):
Automated code review (high effort) — 9 findings on the new PolarDB DynamoDB resources. Findings 1–5 are worth addressing before merge; 1, 2 and 4 fail silently.

alicloud/resource_alicloud_polardb_dynamo.go
```diff
@@ -0,0 +1,732 @@
...
+		)
+		if _, err := stateConf.WaitForState(); err != nil {
+			return WrapErrorf(err, IdMsg, d.Id())
+		}
+	}
+
+	if d.HasChanges("stream_enabled", "stream_view_type") {
+		updateInput := &dynamodb.UpdateTableInput{TableName: aws.String(tableName)}
+		streamSpec := &dynamodb.StreamSpecification{}
+		if v, ok := d.GetOk("stream_enabled"); ok {
```
xuzhang3: **Streams can never be disabled.** `d.GetOk("stream_enabled")` returns `ok=false` for the `false` value, so `streamSpec.StreamEnabled` stays nil. Flipping `stream_enabled` true→false triggers the `HasChanges` block but sends an empty `StreamSpecification`, so the stream is never disabled and every subsequent plan re-shows the diff. Use `d.GetOkExists`/`d.HasChange` + `d.Get` to read the bool unconditionally.

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,614 @@
...
+	}
+
+	// Use PutItem to replace the entire item
+	dynamoItem := make(map[string]*dynamodb.AttributeValue)
+	for k, v := range newAttrs {
+		dynamoItem[k] = jsonToAttributeValue(v)
+	}
+
+	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
+		_, err = dynamoClient.PutItem(&dynamodb.PutItemInput{
```
xuzhang3: **Changing a key value inside `item` orphans the old row.** The key attribute *names* are ForceNew, but their *values* live in the non-ForceNew `item` JSON. Editing e.g. `pk`'s value runs Update → `PutItem` writes a new row and `SetId` is recomputed, but the previous row is never deleted — it leaks in the table outside Terraform's tracking. Either delete the old key first, or mark `item` (or the key values) ForceNew.

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+	tableName := d.Get("table_name").(string)
+	accessKey := d.Get("account_name").(string)
+	secretKey := d.Get("account_auth").(string)
+
+	dynamoClient, err := client.NewPolarDBDynamoClient(endpoint, accessKey, secretKey)
+	if err != nil {
+		return WrapError(err)
+	}
+
+	input := &dynamodb.CreateTableInput{
```
xuzhang3, line 209: **Create ignores `ttl` and `point_in_time_recovery`.** `CreateTableInput` is built without them; they're only applied in Update. A create-time config with `ttl { enabled = true }` or `point_in_time_recovery { enabled = true }` reports apply success while the server has them off, then Read sets them disabled → post-apply diff, and only a second apply converges. Apply them (UpdateTimeToLive / UpdateContinuousBackups) after the table becomes ACTIVE in Create.

alicloud/resource_alicloud_polardb_dynamo.go
```diff
@@ -0,0 +1,732 @@
...
+			},
+			"point_in_time_recovery": {
+				Type: schema.TypeList, Optional: true, Computed: true, MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {Type: schema.TypeBool, Required: true},
+					},
+				},
+			},
+			"server_side_encryption": {
```
xuzhang3: **`server_side_encryption` is completely unwired.** It's declared in the schema but never read, created, or updated — Create doesn't set `SSESpecification`, Read never populates it, Update has no branch. Setting `server_side_encryption { enabled = true }` silently does nothing and the table is created unencrypted with no error or drift.

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,614 @@
...
+		}
+		d.Set("hash_key", hashKey)
+		if rangeKey != "" {
+			d.Set("range_key", rangeKey)
+		}
+	}
+
+	// Build key
+	key := map[string]*dynamodb.AttributeValue{
+		hashKey: extractAttributeValueTyped(hashKeyValue),
```
xuzhang3: **Key type is forced to `S`.** `extractAttributeValueTyped(hashKeyValue)` always builds `{S: ...}` regardless of the real attribute type. For a Number- or Binary-keyed table, GetItem gets a type-mismatched key, returns empty, and Read calls `d.SetId("")` — silently dropping a still-existing item from state (and import of such items always fails). Resolve the key type from the table key schema / attribute definitions.

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+			})
+		}
+		input.AttributeDefinitions = attrs
+	}
+
+	if v, ok := d.GetOk("billing_mode"); ok {
+		input.BillingMode = aws.String(v.(string))
+	}
+
+	if d.Get("billing_mode").(string) == "PROVISIONED" {
```
xuzhang3, line 240: **PROVISIONED default with no capacity fails on first apply.** `billing_mode` defaults to `PROVISIONED` but `read_capacity`/`write_capacity` have no default and `GetOk` skips `0`, so `ProvisionedThroughput` is sent with nil units. A minimal config (hash_key + attribute) hits a CreateTable validation error. Consider defaulting the capacities or requiring them when PROVISIONED.

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+	if err == nil {
+		return false
+	}
+	errMsg := err.Error()
+	return strings.Contains(errMsg, "ResourceNotFoundException") ||
+		strings.Contains(errMsg, "TableNotFound") ||
+		strings.Contains(errMsg, "InvalidDBClusterId.NotFound")
+}
+
+func isDynamoRetryableError(err error) bool {
```
xuzhang3, line 753: **Retry classifier matches over-broad substrings.** `isDynamoRetryableError` matches `"EOF"`, `"dial tcp"`, etc. anywhere in the error text, so a genuinely fatal error whose message merely contains one of these is retried for the full 8-minute window before failing — turning a fast failure into a long hang that masks the real cause.

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,614 @@
...
+	case []interface{}:
+		items := make([]*dynamodb.AttributeValue, 0, len(v))
+		for _, item := range v {
+			items = append(items, jsonToAttributeValue(item))
+		}
+		return &dynamodb.AttributeValue{L: items}
+	case string:
+		return &dynamodb.AttributeValue{S: aws.String(v)}
+	case float64:
+		return &dynamodb.AttributeValue{N: aws.String(fmt.Sprintf("%v", v))}
```
xuzhang3: **Number formatting can produce scientific notation.** The bare `float64` branch uses `fmt.Sprintf("%v", v)`, so `1000000` → `"1e+06"` and `0.0000001` → `"1e-07"`, corrupting the numeric attribute if a raw JSON number reaches this path. Use `strconv.FormatFloat(v, 'f', -1, 64)`.

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,614 @@
...
+func extractAttributeValueString(v interface{}) string {
+	if m, ok := v.(map[string]interface{}); ok {
+		for _, val := range m {
+			return fmt.Sprintf("%v", val)
+		}
+	}
+	return fmt.Sprintf("%v", v)
+}
+
+func keysEqual(key1, key2 map[string]interface{}) bool {
```
xuzhang3, line 650: **Dead code.** `keysEqual` is defined but has zero call sites in the package — remove it.

go.mod
```diff
@@ -85,7 +85,7 @@ require (
 	github.com/apparentlymart/go-cidr v1.1.0 // indirect
 	github.com/apparentlymart/go-textseg/v15 v15.0.0 // indirect
 	github.com/armon/go-radix v1.0.0 // indirect
-	github.com/aws/aws-sdk-go v1.51.8 // indirect
```
xuzhang3, line 88: leave the legacy v1, we can refactor it later. Use v2 AWS SDK for this feature requirement

alicloud/resource_alicloud_polardb_dynamo.go
```diff
@@ -0,0 +1,732 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"endpoint": {
+				Type:        schema.TypeString,
```
xuzhang3: in client.go  the  `NewPolarDBDynamoClient` func assumes this field cannot be empty string, a validate of not empty string should added

alicloud/resource_alicloud_polardb_dynamo.go
```diff
@@ -0,0 +1,732 @@
...
+			"account_auth": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Sensitive:   true,
+				Description: "The authentication password for PolarDB DynamoDB. If not set, the provider's secret key will be used.",
+			},
+			"db_cluster_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
```
xuzhang3: `ValidateFunc: validation.StringIsNotEmpty,`

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The name of the DynamoDB-compatible table.",
+			},
+			"attribute": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Computed:    true,
+				Description: "List of attribute definitions for the table key schema and indexes.",
```
xuzhang3, line 96: The order of the ele in Set depends on the hash, you should add a hash func or use `TypeList` instead.

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+			},
+			"write_capacity": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+			"global_secondary_index": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
```
xuzhang3, line 149: same issue, set should have hash func or user list instead

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+						"read_capacity":      {Type: schema.TypeInt, Optional: true, Computed: true},
+						"write_capacity":     {Type: schema.TypeInt, Optional: true, Computed: true},
+					},
+				},
+			},
+			"local_secondary_index": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
```
xuzhang3, line 176: same

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,732 @@
...
+		}
+		addDebug("CreateTable", output, nil, input)
+		return nil
+	})
+	if err != nil {
+		return WrapErrorf(err, DefaultErrorMsg, "alicloud_polardb_dynamo", "CreateTable", AlibabaCloudSdkGoERROR)
+	}
+
+	dbClusterId := d.Get("db_cluster_id").(string)
+	d.SetId(fmt.Sprintf("%s%s%s", dbClusterId, COLON_SEPARATED, tableName))
```
xuzhang3, line 342: Set ID should after the state sync opt to ensure the resource is in stable state and no errors

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,614 @@
...
+		Read:   resourceAlicloudPolarDBDynamoItemRead,
+		Update: resourceAlicloudPolarDBDynamoItemUpdate,
+		Delete: resourceAlicloudPolarDBDynamoItemDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"endpoint": {
+				Type:        schema.TypeString,
+				Required:    true,
```
xuzhang3: valiation

alicloud/connectivity/client.go
```diff
@@ -398,6 +402,57 @@ func (client *AliyunClient) WithPolarDBClient(do func(*polardb.Client) (interfac
...
+	ak := accessKey
+	sk := secretKey
+	if ak == "" {
+		ak = client.AccessKey
+	}
+	if sk == "" {
+		sk = client.SecretKey
+	}
+
+	log.Printf("[DEBUG] NewPolarDBDynamoClient: endpoint=%s, account_name=%s, account_auth=%s", endpoint, ak, sk)
```
xuzhang3: aksk 别log，容易泄漏

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,709 @@
...
+		}
+		return WrapErrorf(err, DefaultErrorMsg, d.Id(), "DeleteItem", AlibabaCloudSdkGoERROR)
+	}
+
+	return nil
+}
+
+// Helper functions
+
+func buildDynamoItemResourceID(dbClusterId, tableName, hashKey, rangeKey string, attrs map[string]interface{}) string {
```
xuzhang3, line 618: **The resource ID joins raw key values with `:` and is parsed back with an exact part count, so any key value containing a colon makes the item permanently unmanageable.**

`buildDynamoItemResourceID` appends `extractAttributeValueString(hv)` verbatim and joins with `COLON_SEPARATED`, with no escaping. `Read` (line 338-344) then tries `ParseResourceId(d.Id(), 4)` and falls back to `ParseResourceId(d.Id(), 3)` — and `ParseResourceId` enforces an *exact* count (`common.go:961-969`), not a minimum.

With `hash_key = pk`, `range_key = sk` and `item = {"pk":{"S":"user:42"},"sk":{"S":"row1"}}` the ID is `pc-xxx:tbl:user:42:row1` — 5 parts. Both parses fail and `Read` returns `Expected parts' length 3, got 5`. That hits the `Read` at the end of `Create`, every refresh, and the refresh before destroy, so the row exists in the table but the resource can only be removed with `terraform state rm`.

There's a subtler variant: with one colon and no `range_key`, the 4-part parse *succeeds* and silently splits the hash value across `hashKeyValue`/`rangeKeyValue`. That's survivable while `item` is in state (`buildKeyValue` prefers `itemAttrs`), but on `terraform import` `itemAttrs` is nil, so `GetItem` is issued with a truncated key and the item is dropped from state.

The repo already has `ParseResourceIdWithEscaped` (`common.go:979`) and `EscapeColons` (`common.go:1006`) for exactly this case.

alicloud/service_alicloud_polardb.go
```diff
@@ -2341,3 +2350,119 @@ func (s *PolarDBService) DescribeAIDBClusterApiKeys(regionId string) (object map
...
+
+		output, err := dynamoClient.DescribeTable(context.Background(), &dynamodb.DescribeTableInput{
+			TableName: aws.String(tableName),
+		})
+		if err != nil {
+			if isDynamoNotFoundError(err) {
+				return nil, "", nil
+			}
+			if isDynamoRetryableError(err) {
+				return nil, "", err
```
xuzhang3: **Returning the error here defeats the retry it's meant to enable — `WaitForState` aborts on *any* non-nil refresh error.**

`terraform-plugin-sdk@v1.17.2` `helper/resource/state.go:103-113` does:

```go
res, currentState, err := conf.Refresh()
...
if err != nil { resCh <- result; return }
```

There is no tolerance branch — `NotFoundChecks` and `ContinuousTargetOccurence` are only consulted on the `err == nil` paths (state.go:116-173), and `BuildStateConf` (`common.go:1030`) sets no error-tolerating knob. So classifying an error as retryable and then *returning* it produces the opposite of the intended behaviour.

`isDynamoRetryableError` treats any `net.Error` and `io.EOF` as retryable (`resource_alicloud_polardb_dynamo.go:779-783), which is precisely what a just-created public endpoint address produces while DNS propagates. `CreateTable` succeeds, the first poll hits a DNS failure or connection reset, and the entire wait aborts with the table left `CREATING` — the apply fails even though it would have gone `ACTIVE` seconds later.

The retryable branch should return `nil, "", nil` (or a pending status string) so the poll loop continues. Affects the create, update and delete waits at `resource_alicloud_polardb_dynamo.go:358`, `:583` and `:728`.

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,709 @@
...
+	case map[string]interface{}:
+		if len(v) == 1 {
+			for key, val := range v {
+				switch key {
+				case dynamoDataTypeString:
+					return &types.AttributeValueMemberS{Value: fmt.Sprintf("%v", val)}
+				case dynamoDataTypeNumber:
+					return &types.AttributeValueMemberN{Value: formatDynamoNumber(val)}
+				case dynamoDataTypeBinary:
+					return &types.AttributeValueMemberB{Value: []byte(fmt.Sprintf("%v", val))}
```
xuzhang3: **Binary attributes are double-encoded on write and mangled on read, so `B`/`BS` can never round-trip.**

DynamoDB JSON carries `B` values base64-encoded, but this stores the base64 *text* as its literal ASCII bytes with no `base64.StdEncoding.DecodeString`. The SDK then base64-encodes again — `dynamodb@v1.60.1/serializers.go:3557-3559`:

```go
case *types.AttributeValueMemberB:
    av := object.Key("B")
    av.Base64EncodeBytes(uv.Value)
```

So `{"blob":{"B":"SGVsbG8="}}` is stored as `U0dWc2JHOD0`, and any non-Terraform client reading `blob` gets garbage. Same bug for `BS` at lines 96-103, and for the import fallback at line 433 (`[]byte(rawValue)`).

The read path is broken symmetrically: line 141-142 does `string(v.Value)` on the decoded bytes, and the `json.Marshal` at ~line 490 replaces any invalid UTF-8 with U+FFFD. A genuine binary value containing e.g. `0xFF` therefore can never equal the configured `item`, `suppressEquivalentJSONDiffs` can't reconcile it, and every plan shows a diff and re-issues `PutItem`. `BS` has the same issue at lines 167-171.

This isn't a hypothetical surface — `website/docs/r/polardb_dynamo_item.html.markdown:130` advertises `B` and `BS` as supported. Suggest `base64.StdEncoding.DecodeString` on the way in and `base64.StdEncoding.EncodeToString` on the way out.

alicloud/resource_alicloud_polardb_dynamo_item.go
```diff
@@ -0,0 +1,709 @@
...
+	dynamoItem := make(map[string]types.AttributeValue)
+	for k, v := range itemAttrs {
+		dynamoItem[k] = jsonToAttributeValue(v)
+	}
+
+	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
+		_, err = dynamoClient.PutItem(context.Background(), &dynamodb.PutItemInput{
+			TableName:                aws.String(tableName),
+			Item:                     dynamoItem,
+			ConditionExpression:      aws.String("attribute_not_exists(#hk)"),
```
xuzhang3, line 299: **A conditional `PutItem` inside a retry loop, with `d.SetId()` only after the loop, turns a lost response into a permanently orphaned row.**

If the `PutItem` commits server-side but the response is lost, the retry re-issues the *same* conditional write and `attribute_not_exists(#hk)` now fails. `ConditionalCheckFailedException` is not in the retryable set (`resource_alicloud_polardb_dynamo.go:767-771`), so `Create` returns the error at line 323-325 — and `d.SetId(resourceID)` at line 330 is never reached.

The lost-response case is realistic here rather than exotic: `NewPolarDBDynamoClient` sets a hard `Timeout: 15 * time.Second` on the `http.Client` plus `RetryMaxAttempts: 3` (`connectivity/client.go:440-448`), so even a single slow call can produce a `net.Error` on a request that landed — and `isDynamoRetryableError` treats every `net.Error` as retryable (dynamo.go:779-783).

Result: the item is in the table but not in state, every subsequent apply fails with the same conditional error, and `terraform destroy` can't clean it up because the resource was never recorded. Calling `d.SetId()` before the write (or treating `ConditionalCheckFailedException` on a retry attempt as success) would make this recoverable.

alicloud/resource_alicloud_polardb_dynamo.go
```diff
@@ -0,0 +1,933 @@
...
+	if err != nil {
+		return WrapError(err)
+	}
+	dbClusterId := parts[0]
+	tableName := parts[1]
+	endpoint := d.Get("endpoint").(string)
+	accessKey := d.Get("account_name").(string)
+	secretKey := d.Get("account_auth").(string)
+
+	dynamoClient, err := client.NewPolarDBDynamoClient(endpoint, accessKey, secretKey)
```
xuzhang3: **`Update` and `Delete` lack the credential fallback that `Create` and `Read` have, so an empty credential in state silently falls back to the provider AK/SK — which this code says the endpoint rejects.**

`Create` (line 207-215) and `Read` (line 415-422) both resolve from the cluster when the credentials are empty. `Update` (line 548-550) and `Delete` (line 693-695) just do bare `d.Get` and go straight to the client — as do `item.go:513-521` and `item.go:606-614`. `NewPolarDBDynamoClient` then substitutes the provider credentials (`connectivity/client.go:421-429`).

Empty values here are reachable, not theoretical: `account_name`/`account_auth` are `Optional + Computed + Sensitive` and are only persisted inside `Read` (line 420), while `d.SetId()` happens earlier at line 355. If the `ACTIVE` wait at line 365-367 fails — which the state-refresh finding above makes likely — Terraform stores a tainted resource with an ID and *empty* credentials. The next apply's `Delete` signs with the provider AK/SK, which per the comments at `item.go:362` and `service_alicloud_polardb.go:2389` are rejected, so it fails with an opaque signature error. The table is orphaned and the resource can't be destroyed.

Given this block is copy-pasted 8 times, extracting one helper (e.g. `func (s *PolarDBService) dynamoClientFor(d *schema.ResourceData) (*dynamodb.Client, error)`) would fix the divergence and stop it recurring.

alicloud/connectivity/client.go
xuzhang3: **This logs the DynamoDB account password in cleartext, defeating the `Sensitive: true` marking one layer up — and it's on the hot path.**

`sk` is the cluster's `DynamoDBAuthPassword` (`service_alicloud_polardb.go:2426-2431`), and `account_auth` is declared `Sensitive: true` in both new resources (`dynamo.go:71-77`, `item.go:201-207`). `account_name` is Sensitive too. There's no redaction layer anywhere in `alicloud/connectivity/`, so a `TF_LOG=DEBUG terraform apply` — the standard advice when filing a provider bug report — writes the password verbatim into the log file and any CI artifact or issue attachment.

It's also amplified: the client is rebuilt *inside* the state-refresh closure (`service_alicloud_polardb.go:2436-2442`), so a 30-minute create wait at a 3s minimum interval emits the secret hundreds of times in one apply.

Suggest dropping `account_auth` from the format string entirely (the endpoint and account name are the useful debugging bits).

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,933 @@
...
+				gsiConfig["range_key"] = aws.ToString(ks.AttributeName)
+			}
+		}
+		if gsi.Projection != nil {
+			gsiConfig["projection_type"] = string(gsi.Projection.ProjectionType)
+			if len(gsi.Projection.NonKeyAttributes) > 0 {
+				gsiConfig["non_key_attributes"] = gsi.Projection.NonKeyAttributes
+			}
+		}
+		if gsi.ProvisionedThroughput != nil {
```
xuzhang3, line 480: **This guard doesn't protect what the table-level one does: the GSI element is written wholesale, so the omitted capacity keys become `0` in state and the resource never converges.**

Compare with the table-level guard at line 476-479, which skips the `d.Set` calls entirely and so leaves the config value intact. Here the guard only conditionally fills two keys *inside* `gsiConfig`, and then line 501/503 append and `d.Set("global_secondary_index", gsiSet)` unconditionally. `helper/schema`'s object reader fills absent keys with their zero value rather than merging with the previous element, so `read_capacity`/`write_capacity` land in state as `0`.

That the endpoint omits `ProvisionedThroughput` isn't speculation — this PR's own test says so, at `resource_alicloud_polardb_dynamo_test.go:154-156`.

So a GSI configured with `read_capacity = 5, write_capacity = 5` ends up `0/0` in state after apply. `updateDiffDynamoGSI` compares the old/new element maps (line 883-895), sees `capacityChanged`, and sends an `UpdateGlobalSecondaryIndexAction` on *every* apply; the post-apply `Read` resets state to `0` again. Permanent diff plus a real API call each run.

Easiest fix is to mirror the table-level approach: carry the prior element's capacities forward when the response omits them (or omit the whole element rewrite), rather than conditionally setting keys on an element that's written either way.

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,933 @@
...
+	endpoint := d.Get("endpoint").(string)
+	accessKey := d.Get("account_name").(string)
+	secretKey := d.Get("account_auth").(string)
+
+	dynamoClient, err := client.NewPolarDBDynamoClient(endpoint, accessKey, secretKey)
+	if err != nil {
+		return WrapError(err)
+	}
+
+	if d.HasChanges("read_capacity", "write_capacity") && d.Get("billing_mode").(string) == "PROVISIONED" {
```
xuzhang3, line 546: **`Update` has no branch that sends `billing_mode` or `attribute`, and neither is `ForceNew` — so changing them reports a successful apply while the server is untouched.**

`billing_mode` is `Optional` with `Default: "PROVISIONED"` and no `ForceNew` (line 130-135); `BillingMode:` is assigned only in `Create` (line 249-251). `CustomizeDiff` never calls `ForceNew` either — for a non-PROVISIONED plan it returns immediately at line 41. `Update`'s only three mutating branches are this one, `global_secondary_index` (line 593) and `ttl` (line 668).

So switching a table to `PAY_PER_REQUEST` and dropping the capacity attributes runs `Update` with **zero API calls** and returns success at line 686. (The capacities are `Computed`, so their prior values persist and `HasChanges` is false — and this branch requires `PROVISIONED` anyway.) `Read` then re-sets `billing_mode` from the server, so `terraform plan` reports the identical change on every run and the table is never converted.

An `attribute`-only change is the same silent no-op: the attribute set is expanded only inside the GSI-*create* phase (line 646-653, under `if gsiUpdate.Create == nil { continue }`), while `Read` overwrites `attribute` from the server unconditionally at line 471.

Either add an `UpdateTable` branch for `billing_mode`, or mark it (and `attribute`) `ForceNew` if the compatible endpoint can't convert in place.

alicloud/resource_alicloud_polardb_dynamo.go
```diff
@@ -0,0 +1,933 @@
...
+	secretKey := d.Get("account_auth").(string)
+
+	// On import only the resource ID is available, so resolve the DynamoDB-compatible
+	// endpoint address and the DynamoDB account credentials from the cluster to rebuild
+	// the connection info (the endpoint rejects the provider AK/SK fallback).
+	polarDBService := PolarDBService{client}
+	if endpoint == "" {
+		endpoint, err = polarDBService.DescribePolarDBDynamoEndpointAddress(dbClusterId)
+		if err != nil {
+			return WrapError(err)
```
xuzhang3: **`Read` returns a hard error on the endpoint/account pre-steps instead of clearing the ID, so out-of-band deletion of the cluster wedges every future plan *and* destroy.**

The service methods do emit proper NotFound errors — `service_alicloud_polardb.go:2384` and `:2433` both return `WrapErrorf(NotFoundErr(...), NotFoundMsg, ...)` — but `NotFoundError(err)` is never consulted here or at line 418, nor at `item.go:367` / `:374`. The provider convention is:

```go
if err != nil {
    if NotFoundError(err) { d.SetId(""); return nil }
    return WrapError(err)
}
```

(as in `resource_alicloud_polardb_endpoint.go:194-199` and `resource_alicloud_polardb_account.go:172-177`).

Delete the cluster or its DynamoDB-type account outside Terraform and both `terraform plan` and `terraform destroy` fail permanently with a NotFound error — the user has to `terraform state rm` to recover. Worth noting the `DescribeTable` path immediately below *is* handled correctly (line 444-450), so this is only about the two pre-steps.

alicloud/resource_alicloud_polardb_dynamo_table.go
```diff
@@ -0,0 +1,933 @@
...
+	})
+	if err != nil {
+		if isDynamoNotFoundError(err) {
+			d.SetId("")
+			return nil
+		}
+		return WrapErrorf(err, DefaultErrorMsg, d.Id(), "DescribeTable", AlibabaCloudSdkGoERROR)
+	}
+
+	tableDesc := output.Table
```
xuzhang3, line 420: **`output.Table` is dereferenced with no nil check, even though this PR explicitly guards it for the identical call.**

`service_alicloud_polardb.go:2457-2459` has:

```go
if output.Table == nil {
    return nil, "", fmt.Errorf("missing Table field in DescribeTable response")
}
```

so `Table == nil` was evidently considered reachable against this DynamoDB-*compatible* (not real DynamoDB) endpoint. Here there's no such guard between the retry block and line 456's `for _, ks := range tableDesc.KeySchema` — ranging a field of a nil `*types.TableDescription` panics, which crashes the plugin process, so `terraform plan` dies with "plugin crashed" instead of an actionable error and keeps doing so until state is hand-edited.

(The outer `output` pointer is fine — aws-sdk-go-v2 guarantees it non-nil when `err == nil`. It's specifically `.Table` that needs the same guard the refresh func already has.)

xuzhang3 (commented):
Automated review pass over the two new PolarDB DynamoDB resources. 10 findings below (9 confirmed correctness/security, 1 likely panic), each anchored inline.

A few notes that don't anchor cleanly to a single line:

- **`website/docs/r/polardb_dynamo.html.markdown:14` and `polardb_dynamo_item.html.markdown:14` say `Available since v1.286.0`, but v1.286.0 shipped July 24 2026** — `CHANGELOG.md:1` shows the unreleased version is `1.287.0`. Users pinning `1.286.0` will read these docs and get `Invalid resource type`. Should be `v1.287.0`.
- **Resource naming:** `alicloud_polardb_dynamo` manages a DynamoDB *table* (`CreateTable`/`DescribeTable`/`DeleteTable`; the doc's own description says "table resource"; the child resource is `_dynamo_item`). Sibling PolarDB resources are named after the API object (`_endpoint`, `_account`, `_database`), so `alicloud_polardb_dynamo_table` seems right — and a resource type name is frozen public API once released, so this is much cheaper to change now than later. It also leaves room for a future cluster-level DynamoDB resource.
- **`connectivity/client.go:419`** silently rewrites any `https://` endpoint down to `http://` with no opt-out and no warning, so SigV4-signed requests plus all item payloads cross a public endpoint in plaintext. The `strings.Replace(endpoint, "https://", "http://", 1)` is also unanchored (first occurrence anywhere, not just the prefix). Worth at least a warning log, and ideally an escape hatch for when the listener gains TLS.
- **`connectivity/client.go:435`**: `NewStaticCredentialsProvider(ak, sk, "")` drops the STS `SecurityToken` and never calls `config.GetRefreshCredential()` — every neighbouring constructor in that file does both. Under `assume_role` / `ecs_role_name` / `credentials_uri` the provider-credential fallback gets `("", "", "")`, and long applies can't refresh a rotating credential.
- **`resource_alicloud_polardb_account.go:184`**: this file is headed `// Package alicloud. This file is generated automatically. Please do not modify it manually, thank you!` — a hand-edit here will be reverted by the next codegen run, so the fix probably belongs in the describe method or the generator template. The `v.(string)` assertion is also unchecked (the value comes from generic JSON decoding), and the sibling `d.Set` calls in the same function are left unguarded.
- **Acceptance tests**: both `CheckDestroy` funcs `continue`/`return nil` on *every* error, including a failure to build the client and any non-not-found `DescribeTable`/`GetItem` error — so a leaked table or item reports a green destroy check. They also bypass the repo-standard `resourceCheckInitWithDescribeMethod` + `resourceAttrCheckInit` + `rac.checkResourceDestroy()` harness (cf. `resource_alicloud_nas_log_analysis_test.go`), so there's no describe-based existence assertion on create either.
- **Cleanup, in rough priority order**: the endpoint/credential/client bootstrap is copy-pasted across all 8 CRUD entry points and has already diverged (see the Update/Delete finding); `validateDynamoItemJSON` and `suppressEquivalentJSONDiffs` duplicate `validation.StringIsJSON` and `compareJsonTemplateAreEquivalent` (`common.go:750`) — and they're generically-named package-level funcs in the shared `alicloud` package; a fresh `*dynamodb.Client` with its own `&http.Transport{}` is built on every call *and inside the state-refresh closure*, so a 30-minute wait creates hundreds of connection pools that are never closed; `Create` resolves the account then never persists it, so the tail-called `Read` issues the same `DescribeAccounts` again; and `waitDynamoGSIState` hand-rolls a poll loop where `BuildStateConf` (used elsewhere in this same file) would do.

kittenshow98:
image
image

PR #10036, opened 2026-07-24
resource/alicloud_db_instance: support force_delete to destroy Subscription instances

api-tool-agent:
## What & why

Adds a `force_delete` argument to `alicloud_db_instance` (RDS).

By default, Terraform cannot destroy a Prepaid (Subscription) RDS instance — removing the resource from configuration only removes it from state while the cloud instance remains. When `force_delete = true` is set on a Subscription instance, Terraform converts the instance to PostPaid (via `TransformDBInstancePayType`) before running `DeleteDBInstance`, so the Subscription DB instance can be destroyed through Terraform. The remaining subscription period is settled by the cloud.

When `force_delete` is false (default), the existing behavior is preserved.

This mirrors the `force_delete` behavior already available on `alicloud_instance` (ECS).

## Changes
- `alicloud/resource_alicloud_db_instance.go`: new `force_delete` schema field; `resourceAliCloudDBInstanceDelete` now converts Prepaid -> PostPaid before deletion when `force_delete = true` (new `transformDBInstancePayTypeForDelete` helper); preserves the existing WARN + state-only removal when `force_delete = false`.
- `website/docs/r/db_instance.html.markdown`: document `force_delete` and the destroy behavior.
- `alicloud/resource_alicloud_db_instance_test.go`: new acceptance test `TestAccAliCloudRdsDBInstance_Mysql_8_0_PrePaidForceDelete`.

## Acceptance tests

New test (remote ACC verification pending):

```
=== RUN TestAccAliCloudRdsDBInstance_Mysql_8_0_PrePaidForceDelete
```

The test creates a Prepaid MySQL 8.0 RDS instance with `force_delete = true`, then destroys it (expecting Prepaid -> PostPaid conversion followed by DeleteDBInstance). Remote ACC run is in progress.


alicloud/resource_alicloud_db_instance.go
```diff
@@ -75,6 +75,13 @@ func resourceAliCloudDBInstance() *schema.Resource {
 				Optional:         true,
 				DiffSuppressFunc: PostPaidDiffSuppressFunc,
 			},
+			"force_delete": {
+				Type:             schema.TypeBool,
+				Optional:         true,
+				Default:          false,
```
ChenHanZhang: 不需要这个 Default，bool类型默认就是false

alicloud/resource_alicloud_db_instance.go
```diff
@@ -75,6 +75,13 @@ func resourceAliCloudDBInstance() *schema.Resource {
...
+				Description:      "Used to forcibly delete a 'PrePaid' RDS instance. When set to true, the instance is converted to 'PostPaid' before deletion, which settles the remaining subscription period. Set this explicitly when you need to destroy a Subscription DB instance.",
+				DiffSuppressFunc: PostPaidDiffSuppressFunc,
```
ChenHanZhang: 为什么需要这个 DiffSuppressFunc？

alicloud/resource_alicloud_db_instance.go
api-tool-agent: Removed the explicit `Default: false`. As you noted, a bool schema field already defaults to false in the SDK, so the declaration was redundant. The `force_delete` schema now only declares `Type`, `Optional`, and `Description`.

alicloud/resource_alicloud_db_instance.go
api-tool-agent: Removed `DiffSuppressFunc: PostPaidDiffSuppressFunc` from `force_delete`.

`force_delete` is a client-side intent flag: it is not read back from the server (no `d.Set("force_delete", ...)` in Read), and state is preserved from config. It should therefore be shown transparently in `terraform plan`.

Unlike `period`/`auto_renew` (server-echoed attributes on Postpaid instances), `force_delete` carries a billing consequence — setting it converts a PrePaid instance to PostPaid, settling the remaining subscription period. Suppressing a no-op diff on Postpaid instances would hide that intent from the user.

Removing the suppressor is safe:
- In `Delete`, the `PayType == Prepaid` guard runs first, so `force_delete = true` on a Postpaid instance is a no-op.
- Since the field is not written by Read, there is no perpetual diff.

The schema now only declares `Type`, `Optional`, and `Description`.

ChenHanZhang (approved):

ChenHanZhang (approved):

PR #10044, opened 2026-07-27
New Resource: alicloud_ssl_certificates_service_instance_certificate; New Data Source: alicloud_ssl_certificates_service_instance_certificates

api-tool-agent:
### Summary

Adds a read-only view of the certificates issued into a Certificate Management Service (Original
SSL Certificate) instance:

| | |
|---|---|
| `alicloud_ssl_certificates_service_instance_certificate` | Read-only view of a single issued certificate |
| `alicloud_ssl_certificates_service_instance_certificates` | Data source listing issued certificates |

Depends on #10120 — the acceptance tests here build their own certificate and use
`alicloud_ssl_certificates_service_certificate_validation` to wait for issuance.

### Why the certificate is a separate resource from the application

`ApplyCertificate` accepts only an instance ID and returns only a request ID — the certificate it
eventually produces is not in the response. `GetCertificateDetail` is keyed by certificate ID. No
single resource can be keyed by both, so the application and the certificate it yields are separate
resources.

### Notes

- **The resource is read-only.** A certificate is issued by the certificate authority and cannot be
  created through an API, so this adopts one that already exists. Destroying it removes it from
  state and revokes nothing — the certificate goes on existing, and `CheckDestroy` asserts nothing
  about the cloud for that reason.

- **Reissuing produces a new certificate ID.** The documentation calls this out, since a reissue
  changes what this resource refers to.

- **`ListCertificates` reports the certificate ID as a string** while every other API in this
  product returns an integer. The data source normalises it so the declared type holds.

### Acceptance tests

Both cases run in `cn-hangzhou` and build the whole issuance chain — contact, company, instance,
application, wait — so that each owns the certificate it reads, rather than depending on whatever
the account happens to hold.

No domain validation record is published by the tests. The test domain is held in the test account
and the certificate authority validates it without one; publishing a record alongside that races
the validation rather than enabling it.

Acceptance verification is still in progress and results are not being claimed here yet.


website/docs/r/ssl_certificates_service_instance.html.markdown
```diff
@@ -0,0 +1,147 @@
...
+
+Provides a Certificate Management Service (Original SSL Certificate) Instance resource.
+
+A certificate instance is the subscription that a certificate is issued against. Creating this resource
+places a subscription order and then writes the certificate application details, such as the bound domain
+name, the CSR and the contacts, onto the instance.
+
+For information about Certificate Management Service (Original SSL Certificate) Instance and how to use it, see [What is Instance](https://next.api.alibabacloud.com/document/BssOpenApi/2017-12-14/CreateInstance).
+
+-> **NOTE:** Available since v1.284.0.
```
LCLSpring: 这个版本号不对。

website/docs/d/ssl_certificates_service_instances.html.markdown
```diff
@@ -0,0 +1,103 @@
...
+sidebar_current: "docs-alicloud-datasource-ssl-certificates-service-instances"
+description: |-
+  Provides a list of Ssl Certificates Service Instance owned by an Alibaba Cloud account.
+---
+
+# alicloud_ssl_certificates_service_instances
+
+This data source provides Ssl Certificates Service Instance available to the user.[What is Instance](https://next.api.alibabacloud.com/document/BssOpenApi/2017-12-14/CreateInstance)
+
+-> **NOTE:** Available since v1.284.0.
```
LCLSpring: 同样地，版本号要对齐下一个已发布版本

api-tool-agent:
Split: the certificate_validation resource moved to #10120 and the instance_certificate resource + data source moved to #10121 (with tests redesigned to adopt existing issued certificates instead of issuing their own). Nothing remains here.

PR #10085, opened 2026-07-30
resource/alicloud_ecs_disk: defer BurstingEnabled until cloud_auto on ESSD->AutoPL

api-tool-agent:
## Why

When `category` is changed to `cloud_auto` and `bursting_enabled` is set in the same apply, the disk is still on its previous category (e.g. `cloud_essd`) when `ModifyDiskAttribute` runs, so `BurstingEnabled` is rejected by the OpenAPI:

> An error is reported if you specify this parameter for a disk that does not support performance burst.

This forced the ESSD -> AutoPL migration plus enabling the performance burst feature to be split across two applies.

## What

Defer the `BurstingEnabled` API call to after `ModifyDiskSpec` + `WaitForState` when `category` is being changed to `cloud_auto` in the same apply, so the disk is already `cloud_auto` when `BurstingEnabled` is set.

The reverse direction (`cloud_auto` -> non-`cloud_auto`) stays safe because `BurstingEnabled` is still set inside `ModifyDiskAttribute` while the disk remains `cloud_auto` at that point.

## Tests

New Update test covers the migration path in a single apply:

- Step 1: create a `cloud_essd` disk
- Step 2: change `category` to `cloud_auto`, enable `bursting_enabled` and set `provisioned_iops` in a single apply
- Step 3: import verify

```
=== RUN TestAccAliCloudECSDisk_basic8019_essd_to_autopl_burst
```

### DIFF
--- a/alicloud/resource_alicloud_ecs_disk.go
+++ b/alicloud/resource_alicloud_ecs_disk.go
@@ -403,6 +403,11 @@ func resourceAliCloudEcsDiskUpdate(d *schema.ResourceData, meta interface{}) err
 	var response map[string]interface{}
 	var query map[string]interface{}
 	update := false
+	// needBurstingPost defers the BurstingEnabled API call to after ModifyDiskSpec +
+	// WaitForState when category is being changed to cloud_auto in the same apply.
+	// BurstingEnabled is only valid on cloud_auto disks (OpenAPI returns an error if
+	// specified on a disk that does not support performance burst).
+	needBurstingPost := false
 	d.Partial(true)
 
 	action := "ModifyDiskAttribute"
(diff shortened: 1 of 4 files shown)


ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

ChenHanZhang (approved):

api-tool-agent:
Closing in favor of a fresh PR. This branch went stale and could no longer trigger the AccTest workflow on CI. Re-submitting the same verified fix on a new branch.

PR #10113, opened 2026-08-03
New Resource: alicloud_ssl_certificates_service_instance; New Data Source: alicloud_ssl_certificates_service_instances

api-tool-agent:
Adds the SSL Certificates Service certificate instance resource and its data source.

An instance is the purchased entitlement a certificate is later issued against, so it carries the application configuration (`domain`, `csr`, `validation_method`, `key_algorithm`, `generate_csr_method`) alongside the order parameters.

### Notes on a few fields

- `city` / `province` are read-only. The server derives the address from the associated company, or applies a default when no company is set, because certificate authorities reject non-standard addresses.
- `domain` and `csr` are `Optional + Computed`. Which one wins depends on `generate_csr_method`: with `upload` the domain is taken from the CSR's common name, with `online` the supplied value is used.
- `contact_id_list` is a `TypeSet`. The server does not preserve the submitted order.

### Delete

A subscription still within its period cannot be deleted outright, so the instance is refunded first and the refund's failure tolerated — an order that is no longer refundable is deletable as it stands.

An instance with a certificate application in flight sits in `pending`, where it can be neither refunded nor deleted. Delete waits for it to settle rather than withdrawing the application, since withdrawing an application the configuration never declared is not this resource's decision.

### Test

`TestAccAliCloudSslCertificatesServiceInstance_basic` and `TestAccAliCloudSslCertificatesServiceInstanceDataSource` pass in `cn-hangzhou`.


alicloud/resource_alicloud_ssl_certificates_service_instance.go
```diff
@@ -0,0 +1,602 @@
+// Package alicloud. This file is generated automatically. Please do not modify it manually, thank you!
```
LCLSpring: 这个资源有不少手写逻辑，不是自动生成的。

website/docs/r/ssl_certificates_service_instance.html.markdown
```diff
@@ -0,0 +1,172 @@
...
+
+# alicloud_ssl_certificates_service_instance
+
+Provides a Certificate Management Service (Original SSL Certificate) Instance resource.
+
+
+
+For information about Certificate Management Service (Original SSL Certificate) Instance and how to use it, see [What is Instance](https://next.api.alibabacloud.com/document/BssOpenApi/2017-12-14/CreateInstance).
+
+-> **NOTE:** Available since v1.287.0.
```
LCLSpring, line 17: 这个版本号注意更新

website/docs/r/ssl_certificates_service_instance.html.markdown
```diff
@@ -0,0 +1,172 @@
...
+
+* `company_id` - (Optional, Computed) The company information ID.
+
+~> **NOTE:** The address recorded on the instance — `city` and `province` — comes from the company referenced by `company_id`, and from a server-side default when no company is referenced. Certificate authorities reject non-standard addresses, so the address is taken from the company record rather than accepted per instance; set it on `alicloud_ssl_certificates_service_company` instead. Both attributes are read-only here.
+ This parameter is required for OV and EV certificates. Otherwise, you cannot call the ApplyCertificate operation to apply for a certificate.
+* `contact_id_list` - (Optional, Set) The list of contact IDs. At least one contact ID is required. Otherwise, you cannot call the ApplyCertificate operation to apply for a certificate.
```
LCLSpring: 这里在说API，从tf的角度，应该是通过 certificate apply 资源去申请证书。文档中应该用 certificate apply 资源 去提醒用户

website/docs/d/ssl_certificates_service_instances.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+sidebar_current: "docs-alicloud-datasource-ssl-certificates-service-instances"
+description: |-
+  Provides a list of Ssl Certificates Service Instance owned by an Alibaba Cloud account.
+---
+
+# alicloud_ssl_certificates_service_instances
+
+This data source provides Ssl Certificates Service Instance available to the user.[What is Instance](https://next.api.alibabacloud.com/document/BssOpenApi/2017-12-14/CreateInstance)
+
+-> **NOTE:** Available since v1.287.0.
```
LCLSpring, line 14: 注意版本号，要写下个将会发布的版本

PR #10115, opened 2026-08-03
New Resource: alicloud_ssl_certificates_service_certificate_apply; resource/alicloud_ssl_certificates_service_instance: Fixed the refund failure when deleting a freshly created instance

api-tool-agent:
Adds `alicloud_ssl_certificates_service_certificate_apply`, which submits a certificate application against a purchased certificate instance and reports the domain ownership validation records the certificate authority expects to be published.

The application is a distinct object from both the instance that pays for it and the certificate it eventually produces: `ApplyCertificate` takes only an instance id and returns no certificate id, while the certificate is keyed separately. Modelling it as its own resource is what allows a configuration to express "apply, then publish these records, then wait", and what allows an application to be withdrawn on destroy without touching the instance or revoking anything.

Notes on the design:

- The application configuration fields (`domain`, `csr`, `validation_method`, `key_algorithm`, `generate_csr_method`) are declared on this resource in addition to the instance. `ApplyCertificate` accepts only the instance id, so without them a change to the instance's application configuration would produce no plan here and the certificate would never be re-applied for. They are `ForceNew` and validated against the instance before the call, so a configuration that sets them inconsistently fails with an explanation rather than silently applying for something else.
- Create waits for the validation records to appear and then hold across consecutive reads: an application is not operable the instant its records are published, and returning earlier hands back a resource that cannot yet be destroyed.
- Destroy calls `CancelPendingCertificate` and waits for the instance to leave the pending state, because a pending instance can be neither refunded nor deleted. Revoking an already-issued certificate is deliberately not done: the certificate is a separate object that other services may be serving, and revocation is irreversible.
- The acceptance test applies for a domain the account does not hold, so the application stays outstanding for the duration. A domain the account does hold is validated by the certificate authority on its own within minutes, which would settle the application before the validation records or the withdrawal could be exercised.

Also fixes a refund race in `alicloud_ssl_certificates_service_instance`: a freshly purchased instance answers `RefundInstance` with `CouldNotRefund.NotSupport` for the first seconds of its life while the purchase order is still settling, which left an instance destroyed right after creation unrefunded and its deletion stuck behind delete protection. The refund is now retried briefly before the refusal is taken at its word; instances that genuinely cannot be refunded keep the existing tolerant behaviour.

### DIFF
--- a/alicloud/resource_alicloud_ssl_certificates_service_certificate_apply.go
+++ b/alicloud/resource_alicloud_ssl_certificates_service_certificate_apply.go
@@ -1,0 +1,352 @@
+package alicloud
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aliyun/terraform-provider-alicloud/alicloud/connectivity"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceAliCloudSslCertificatesServiceCertificateApply() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAliCloudSslCertificatesServiceCertificateApplyCreate,
+		Read:   resourceAliCloudSslCertificatesServiceCertificateApplyRead,
+		Delete: resourceAliCloudSslCertificatesServiceCertificateApplyDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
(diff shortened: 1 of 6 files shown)


alicloud/resource_alicloud_ssl_certificates_service_certificate_apply.go
```diff
@@ -0,0 +1,342 @@
+// Package alicloud. This file is generated automatically. Please do not modify it manually, thank you!
```
LCLSpring: 这是个手写资源，不用这行注释标识

api-tool-agent: Removed. The generator header was misleading here: the create waits on the validation records, the configuration is checked against the instance before submitting, and the delete has its own withdrawal handling — none of which the generator produces.

PR #10120, opened 2026-08-03
New Resource: alicloud_ssl_certificates_service_certificate_validation

api-tool-agent:
### Summary

Adds `alicloud_ssl_certificates_service_certificate_validation` for Certificate Management Service
(Original SSL Certificate).

`alicloud_ssl_certificates_service_certificate_apply` submits an application and reports the domain
validation records the certificate authority wants published, but issuance happens on the authority's
side and takes minutes. This resource is the wait in between: it blocks until the certificate reaches
`issued`, so anything that consumes the certificate can depend on it.

### Notes

- **It calls no API of its own.** Create polls `GetInstanceDetail` until `CertificateStatus` becomes
  `issued`; `revoked` and `expired` are treated as failure states. Destroying it does nothing — the
  certificate it waited for goes on existing, so the resource is simply dropped from state.

- **`validation_record_ids` maps to no API field.** It exists so a configuration can order the wait
  after whatever publishes the domain validation records, in the same way `depends_on` would. It is
  never read back, which is why import verification is off for this resource.

- **Create is given a 75-minute timeout.** Issuance time is set by the certificate authority, not by
  the API, and DV issuance has been observed taking upwards of ten minutes.

### Acceptance tests

Two cases in `cn-hangzhou`, one per outcome of a wait:

- `_basic` — the success path. It adopts an instance whose certificate has already been issued,
  found through the `alicloud_ssl_certificates_service_instances` data source, and asserts the wait
  completes and reports the certificate. Adopting rather than issuing is deliberate: an instance
  whose certificate has been issued can be neither refunded nor deleted, so a test that issued its
  own certificate would permanently leave a paid instance behind on every run.

- `_unissuable` — the failure path. It applies for a certificate on a domain the account does not
  hold, so validation never succeeds, and asserts the wait times out loudly rather than reporting
  success. Because nothing is ever issued, the application can be withdrawn and the instance
  refunded — the destroy step tears the whole chain down and doubles as the cleanup assertion.

### DIFF
--- a/alicloud/resource_alicloud_ssl_certificates_service_certificate_validation.go
+++ b/alicloud/resource_alicloud_ssl_certificates_service_certificate_validation.go
@@ -1,0 +1,116 @@
+// This file is maintained by hand. The certificate validation resource is a wait with no backing
+// API of its own — something the generator cannot express — so unlike its sibling files it is not
+// regenerated from a specification.
+package alicloud
+
+import (
+	"log"
+	"time"
+
+	"github.com/aliyun/terraform-provider-alicloud/alicloud/connectivity"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceAliCloudSslCertificatesServiceCertificateValidation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAliCloudSslCertificatesServiceCertificateValidationCreate,
+		Read:   resourceAliCloudSslCertificatesServiceCertificateValidationRead,
+		Delete: resourceAliCloudSslCertificatesServiceCertificateValidationDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			// Domain validation and issuance are performed by the certificate authority. A DV
+			// certificate is usually issued within minutes of the validation records going live,
+			// while OV and EV certificates additionally go through a manual review that can take
+			// hours. Raise this timeout accordingly for those.
+			Create: schema.DefaultTimeout(75 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"instance_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			// The IDs of whatever published the domain ownership validation records — typically
+			// alicloud_alidns_record resources created from the domain_validation_list reported by
+			// alicloud_ssl_certificates_service_certificate_apply. The values map to no API field
+			// and are never read; the attribute exists so that referencing the record IDs here
+			// makes Terraform publish the records before this wait starts, the same way an explicit
+			// depends_on would.
+			"validation_record_ids": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"certificate_id": {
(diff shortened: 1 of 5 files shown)


website/docs/r/ssl_certificates_service_certificate_validation.html.markdown
```diff
@@ -0,0 +1,122 @@
...
+  Provides a Alicloud Certificate Management Service (Original SSL Certificate) Certificate Validation resource.
+---
+
+# alicloud_ssl_certificates_service_certificate_validation
+
+Provides a Certificate Management Service (Original SSL Certificate) Certificate Validation resource.
+
+Waits until the certificate application submitted on a certificate instance has been issued, and exposes the resulting certificate for downstream services to reference.
+
+For information about Certificate Management Service (Original SSL Certificate) Certificate Validation and how to use it, see [What is Certificate Validation](https://next.api.alibabacloud.com/document/cas/2020-04-07/GetInstanceDetail).
```
LCLSpring: 你给出的链接是一个API的链接，用户通过这个链接，能明白什么是证书校验么？

alicloud/resource_alicloud_ssl_certificates_service_certificate_validation.go
```diff
@@ -0,0 +1,108 @@
+// Package alicloud. This file is generated automatically. Please do not modify it manually, thank you!
```
LCLSpring: 这是个手写资源

alicloud/resource_alicloud_ssl_certificates_service_certificate_validation.go
```diff
@@ -0,0 +1,108 @@
...
+			Create: schema.DefaultTimeout(75 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"instance_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"validation_record_ids": {
```
LCLSpring, line 42: 解释一下这个属性用来做什么的？来自哪里？

website/docs/r/ssl_certificates_service_certificate_validation.html.markdown
api-tool-agent: Fair point — an API reference explains nothing about the concept. Fixed in 237248dda: the doc now explains inline what validation is (the certificate authority verifying domain control via the records from `domain_validation_list`, with issuance following), and the link now points to the product help page — the same one the merged `ssl_certificates_service_certificate` doc uses — instead of the GetInstanceDetail API page.

alicloud/resource_alicloud_ssl_certificates_service_certificate_validation.go
api-tool-agent: Right — the "generated automatically" header was wrong for this file. Fixed in 237248dda: the header now states the file is maintained by hand, and why — the resource is a wait with no backing API of its own, which the generator cannot express.

alicloud/resource_alicloud_ssl_certificates_service_certificate_validation.go
api-tool-agent: Documented in 237248dda, right above the schema field. In short: the values come from whatever published the domain validation records — typically `alicloud_alidns_record` resources created from `certificate_apply`'s `domain_validation_list`. They map to no API field and are never read; referencing the record IDs here makes Terraform publish the records before the wait starts, the same way an explicit `depends_on` would. That is also why import verification skips it — an imported wait cannot recover a value that exists only in configuration.

PR #10130, opened 2026-08-04
resource/alicloud_cr_ee_instance: Added the field tags

data-source/alicloud_cr_ee_instances: Added the field tags

testcase/alicloud_cr_ee_instance: Fixed the RAM policy scope of the custom OSS bucket case

api-tool-agent:
Add tag support to the CR Enterprise Edition instance resource and data source, backed by the cr `TagResources` / `UntagResources` / `ListTagResources` APIs.

### resource/alicloud_cr_ee_instance

- New `tags` argument.
- Read picks tags up from the existing `GetInstance` response, which already returns `Tags`, so no extra API call is added to the read path.
- Update applies the diff through `CrServiceV2.SetResourceTags` (`TagResources` / `UntagResources`, `ResourceType` `Instance`). Create reaches tags through the existing Create -> Update chain.
- The `acs:` system tags CR attaches on its own (such as `acs:rm:rgId`) are filtered by the existing `tagsToMap` / `ignoredTags` helpers, so they do not produce a perpetual diff.

### datasource/alicloud_cr_ee_instances

- New `tags` filter: an instance is returned only when it carries every requested tag with the same value.
- New per-instance `tags` attribute, read through `CrServiceV2.ListCrInstanceTags` with `NextToken` pagination.

### testcase/alicloud_cr_ee_instance

`TestAccAliCloudCrInstance_basic7970_modified` has been failing on a 6m create timeout. The cause is in the test itself: it creates the RAM policy behind `AliyunContainerRegistryCustomizedOSSBucketRole` only when no policy of that name exists, and reuses an existing one without checking or updating its content, while the policy document scoped the OSS permissions to `${var.name}` — the bucket of whichever run created it first. Since every run generates a new randomised bucket name, from the second run onwards CR held only `ListBuckets` on the bucket passed as `custom_oss_bucket`, instance activation never completed and the order was never committed.

`tf-testacc-*` and `tfacc*` wildcards are added alongside the existing entries so a freshly created policy covers every bucket name the CR test functions in this file can generate, with a comment recording the constraint.

### Acceptance tests

All eight cases pass, including the previously failing one:

```
TestAccAliCloudCrInstance_tags                    PASS  185.49s   (new)
TestAccAlicloudCREEInstancesDataSource            PASS  186.80s
TestAccAliCloudCrInstance_Basic                   PASS  134.14s
TestAccAliCloudCrInstance_Standard                PASS  192.07s
TestAccAliCloudCrInstance_Advanced                PASS  169.69s
TestAccAliCloudCrInstance_EconomyDisableScanner   PASS  168.25s
TestAccAliCloudCrInstance_basic8613               PASS  261.95s
TestAccAliCloudCrInstance_basic7970_modified      PASS  226.84s   (was a 372s timeout failure)
```

`TestAccAliCloudCrInstance_tags` covers the full tag lifecycle: create with tags, change values and add one, remove a subset, clear all, then import.

The existing cases were re-run to check for regressions. Their debug logs contain no `TagResources` / `UntagResources` / `ListTagResources` calls at all, confirming the new branch stays inert when `tags` is not configured.

### DIFF
--- a/alicloud/data_source_alicloud_cr_ee_instances.go
+++ b/alicloud/data_source_alicloud_cr_ee_instances.go
@@ -1,6 +1,7 @@
 package alicloud
 
 import (
+	"fmt"
 	"regexp"
 	"sort"
 
@@ -28,6 +29,7 @@ func dataSourceAlicloudCrEEInstances() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"tags": tagsSchema(),
 
 			// Computed values
 			"ids": {
@@ -96,6 +98,11 @@ func dataSourceAlicloudCrEEInstances() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+						"tags": {
+							Type:     schema.TypeMap,
+							Computed: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
 					},
 				},
 			},
@@ -106,6 +113,7 @@ func dataSourceAlicloudCrEEInstances() *schema.Resource {
 func dataSourceAlicloudCrEEInstancesRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*connectivity.AliyunClient)
 	crService := &CrService{client}
+	crServiceV2 := CrServiceV2{client}
 	pageNo := 1
 	pageSize := 50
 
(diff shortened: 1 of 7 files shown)


api-tool-agent:
Note on the failing `TestingCoverageRate` check.

The job reports 82.35% coverage with three attributes missing test cases: `password`, `kms_encrypted_password` and `kms_encryption_context`. This is pre-existing debt rather than something this PR introduces — the check only runs when the resource is touched, and the steps covering those three attributes have been commented out in `TestAccAliCloudCrInstance_Basic` since long before this change, with the note "Currently, the API does not support sts".

I tried enabling them and confirmed the note is still accurate. The `cr` `ResetLoginPassword` API rejects the credentials used by the acceptance test framework:

```
Code: STS_NOT_SUPPORT
Message: STS is not supported.
RequestId: 019FCBFA-FBDA-54DA-A088-BDF2D5779CF2
HostId: cr.cn-beijing.aliyuncs.com
```

The step fails during apply, so the three attributes cannot be exercised. I reverted that attempt and instead recorded the error code and RequestId in the comment above the disabled steps, so the reason is verifiable next time.

Worth noting that `TestingCoverageRate` is a static check over the test configs, so simply writing `password` into a config block would turn the job green while leaving the acceptance test broken. I did not want to do that. The `modified coverage rate` reported by the same job is already 100%, i.e. the `tags` attribute this PR adds is fully covered.

Please advise how you would like this handled — whether the three attributes should be exempted somehow, or the check tolerated for this resource until the API supports the test credentials.

Issue #10152, opened 2026-08-06
New Data Source: alicloud_sae_application_instances (expose SAE instance ENI IPs)


Explorer1092:
## Feature request

There is currently **no way to obtain the ENI IP addresses of SAE application instances** from this provider:

- `alicloud_sae_application` (resource) has no computed attribute for instance IPs.
- `alicloud_sae_applications` (data source) supports *filtering by* `instanceIps` (`field_type`), but does not *export* instance IPs.
- `SaeService` has no wrapper for `DescribeApplicationInstances`.

The underlying APIs exist and are stable:
- `GET /pop/v1/sam/app/describeApplicationGroups` (AppId → GroupIds)
- `GET /pop/v1/sam/app/describeApplicationInstances` (AppId + GroupId → instances with `InstanceContainerIp`, `InstanceContainerStatus`, `InstanceHealthStatus`, `ImageUrl`, `PackageVersion`, `VSwitchId`, ...)

## Use case

SAE **reassigns ENI IPs on every rolling deployment** (`DeployApplication`, `RestartApplication`). Any downstream resource that points at pod IPs — e.g. a self-managed ALB server group (`alicloud_alb_server_group` with `server_group_type = "Ip"`) fronting an SAE app — goes stale after each deploy and must be updated manually. (SAE-managed CLB ingress follows ENIs automatically, but it does not expose the listener timeouts we need, e.g. `request_timeout`/`idle_timeout` = 600s for long-lived streaming responses, so a self-managed ALB is required.)

With an `alicloud_sae_application_instances` data source, the server group can be wired declaratively:

```terraform
data "alicloud_sae_application_instances" "app" {
  application_id = alicloud_sae_application.app.id
  # read after the rolling deployment finishes
  depends_on     = [alicloud_sae_application.app]
}

resource "alicloud_alb_server_group" "app" {
  server_group_type = "Ip"
  dynamic "servers" {
    for_each = toset(data.alicloud_sae_application_instances.app.instances.*.instance_container_ip)
    content {
      server_id   = servers.value
      server_ip   = servers.value
      server_type = "Ip"
      port        = 4000
    }
  }
  # ...
}
```

Since the `alicloud_sae_application` update with `deploy = true` already waits for the ChangeOrder to reach a terminal state, reading instance IPs after the resource update yields the post-deploy IPs within a single `terraform apply` — no manual step, no stale backends.

## Proposed schema

[...]

PR #10153, opened 2026-08-06
New Data Source: alicloud_sae_application_instances

Explorer1092:
Implements #10152.

## What

Adds a new data source `alicloud_sae_application_instances` that lists the instances of an SAE application, exposing per-instance ENI IP (`instance_container_ip`), container/health status, `image_url`, `package_version` and `vswitch_id`.

Implementation:
- `group_id` unset → all groups are enumerated via `GET /pop/v1/sam/app/describeApplicationGroups`; `group_id` set → only that group.
- Instances are read via `GET /pop/v1/sam/app/describeApplicationInstances` with pagination (`PageSize`/`CurrentPage`) and the standard retry wrapper.
- Optional client-side `ids` filtering, `output_file`, and `ids` export, following the conventions of `alicloud_sae_applications`.
- Documentation page + sidebar entry (website/alicloud.erb) + acceptance test included.

## Why

SAE reassigns ENI IPs on every rolling deployment. Downstream IP-based resources (e.g. a self-managed `alicloud_alb_server_group` fronting an SAE app, needed when listener timeouts beyond the SAE-managed CLB defaults are required) currently have no way to track the new IPs declaratively. Because the `alicloud_sae_application` update with `deploy = true` already waits for the ChangeOrder to complete, `depends_on` lets the data source read the post-deploy IPs within a single apply:

```terraform
data "alicloud_sae_application_instances" "app" {
  application_id = alicloud_sae_application.app.id
  depends_on     = [alicloud_sae_application.app]
}

resource "alicloud_alb_server_group" "app" {
  server_group_type = "Ip"
  dynamic "servers" {
    for_each = toset(data.alicloud_sae_application_instances.app.instances.*.instance_container_ip)
    content {
      server_id   = servers.value
      server_ip   = servers.value
      server_type = "Ip"
      port        = 4000
    }
  }
}
```

## Testing

- `go build ./alicloud/` and `go vet ./alicloud/` pass (the two `fmt.Sprintln` vet notes in `alicloud/common.go` are pre-existing upstream).
- Acceptance test `TestAccAlicloudSAEApplicationInstancesDataSource` added following the existing SAE data source test pattern (exist: application_id; fake: ids filter). It requires an Alibaba Cloud account to run, so it has not been executed here — happy to adjust if maintainers hit issues running it.

### DIFF
--- a/website/docs/d/sae_application_instances.html.markdown
+++ b/website/docs/d/sae_application_instances.html.markdown
@@ -1,0 +1,63 @@
+---
+subcategory: "Serverless App Engine (SAE)"
+layout: "alicloud"
+page_title: "Alicloud: alicloud_sae_application_instances"
+sidebar_current: "docs-alicloud-datasource-sae-application-instances"
+description: |-
+  Provides a list of SAE application instances to the user.
+---
+
+# alicloud\_sae\_application\_instances
+
+This data source provides the instances of an SAE application, including the ENI IP (`instance_container_ip`) of each instance.
+
+-> **NOTE:** Available in v1.289.0+.
+
+## Example Usage
+
+Basic Usage
+
+```terraform
+data "alicloud_sae_application_instances" "default" {
+  application_id = "47a14017-2587-4db1-80f1-3285453f7459"
+}
+
+output "sae_instance_ips" {
+  value = data.alicloud_sae_application_instances.default.instances.*.instance_container_ip
+}
+```
+
+Filter by instance ids:
+
+```terraform
+data "alicloud_sae_application_instances" "default" {
+  application_id = "47a14017-2587-4db1-80f1-3285453f7459"
+  ids            = ["ai-gateway-litellm-test-47a14017-2587-4db1-3285-qc4rh"]
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `application_id` - (Required) The ID of the SAE application.
+* `group_id` - (Optional) The ID of the instance group. If not set, instances of all groups of the application are returned.
+* `ids` - (Optional) A list of instance IDs to filter results.
+* `output_file` - (Optional) File name where to save data source results (after running `terraform plan`).
+
+## Attributes Reference
+
+The following attributes are exported in addition to the arguments listed above:
+
+* `ids` - A list of instance IDs.
+* `instances` - A list of instances. Each element contains the following attributes:
+    * `id` - The ID of the instance.
+    * `instance_id` - The ID of the instance.
+    * `group_id` - The ID of the group the instance belongs to.
+    * `instance_container_ip` - The ENI IP address of the instance container.
+    * `instance_container_status` - The status of the instance container.
+    * `instance_health_status` - The health status of the instance.
+    * `main_container_status` - The status of the main container.
+    * `image_url` - The image address the instance is running.
+    * `package_version` - The package version the instance is running.
+    * `vswitch_id` - The vSwitch ID of the ENI of the instance.
--- a/alicloud/data_source_alicloud_sae_application_instances.go
+++ b/alicloud/data_source_alicloud_sae_application_instances.go
@@ -1,0 +1,232 @@
+package alicloud
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/PaesslerAG/jsonpath"
+	"github.com/aliyun/terraform-provider-alicloud/alicloud/connectivity"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceAlicloudSaeApplicationInstances() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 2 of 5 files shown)


xuzhang3 (approved):

website/docs/d/sae_application_instances.html.markdown
```diff
@@ -0,0 +1,63 @@
+---
+subcategory: "Serverless App Engine (SAE)"
+layout: "alicloud"
+page_title: "Alicloud: alicloud_sae_application_instances"
+sidebar_current: "docs-alicloud-datasource-sae-application-instances"
+description: |-
+  Provides a list of SAE application instances to the user.
+---
+
+# alicloud\_sae\_application\_instances
```
xuzhang3, line 10: no need to add the `\` any more.  can be removed

xuzhang3:
Thanks for the thorough writeup — the use case (ENI IPs churning on every rolling deploy, self-managed ALB server group needing them) is clear, and the issue-then-PR flow is appreciated. Build and vet are clean on my side too; the two `fmt.Sprintln` notes in `common.go` are indeed pre-existing.

A few things to fix before this can go in.

### 1. `document-check` CI will fail on the doc page

Running the repo's own checker against the new page:

```
==> Checking docs content of website/docs/d/sae_application_instances.html.markdown ...
[WARNING] line 10: please remove the \
[Error] line 14: Expected: -> **NOTE:** Available since v1.289.0. Got: -> **NOTE:** Available in v1.289.0+.
exit status 1
```

So: drop the escaping in the `# alicloud_sae_application_instances` heading (as I noted inline — the surrounding older pages still have it, but the checker no longer accepts it), and use the `Available since v1.289.0.` wording. `v1.289.0` is the right target version.

### 2. Hard error when a group has no instances

```go
resp, err := jsonpath.Get("$.Data.Instances", response)
if err != nil {
    return WrapErrorf(err, FailedGetAttributeMsg, action, "$.Data.Instances", response)
}
```

When a group has zero instances, SAE returns `Data` with only `CurrentPage`/`TotalSize`/`PageSize` and no `Instances` key, and `jsonpath` fails with `unknown key Instances`; a `"Data": null` body fails with `unsupported value type <nil> for select`. Either way `terraform plan` errors out instead of returning an empty list. That is a reachable state for a stopped application, an application scaled to 0 replicas, or — with `group_id` unset — any one of several groups whose deploy has not produced pods yet, which then fails the whole read.

The current convention in newer data sources is to discard the error and let the type assertion absorb the nil:

```go
resp, _ := jsonpath.Get("$.Data.Instances", response)
result, _ := resp.([]interface{})
```

`result, _ := resp.([]interface{})` below already handles `nil` correctly, so this is a one-line change. Same treatment for `$.Data` in the groups loop.

### 3. Doc examples must be self-contained

Both `Example Usage` blocks hard-code a literal `application_id` (and a literal instance ID in the `ids` example). Two problems:

- The docs-example CI job actually runs `terraform` against the example blocks, and a data source pointing at an application ID that does not exist in the test account will fail there.
- Those values look like they came from a live account. Please replace them with a self-contained example that creates the namespace + application and references `alicloud_sae_application.default.id`, the way `website/docs/d/sae_applications.html.markdown` does. If you want to keep a literal for illustration, mask the tail with `****`.

### 4. Naming convention

New files use the `AliCloud` spelling — `dataSourceAliCloudSaeApplicationInstances` / `dataSourceAliCloudSaeApplicationInstancesRead`. The `Alicloud` form here matches the older neighbouring SAE files, but new code should use the current one.

### 5. Acceptance test

- `image_url` is pinned to `registry-vpc.cn-hangzhou.aliyuncs.com/...` while the namespace is built from `defaultRegionToTest` and `checkoutSupportedRegions` admits every SAE region — outside cn-hangzhou the VPC image pull fails. `resource_alicloud_sae_application_test.go` templates it: `fmt.Sprintf("registry-vpc.%s.aliyuncs.com/...", defaultRegionToTest)`. (`sae_applications_test.go`, which this was modelled on, has the same latent problem.)
- The exist check pins `ids.# = 2` and asserts `CHECKSET` on `instance_container_ip`. Create waits for the change order to reach `{2, 8, 11, 12}`, and 8/11/12 are non-success terminals, so pods may not have IPs yet on an unlucky run. Worth confirming when the test runs.

No worries about not having run it — I will run `TestAccAlicloudSAEApplicationInstancesDataSource` and post the result here.

### Optional, but easier now than later

A data source schema is awkward to extend semantically once released, so consider adding while you are here:

- A `status` filter on `instance_container_status`. Your own use case wants only `Running` pods in the ALB server group — right now `Pending`/`Terminating` IPs get fed in too, which is the failure mode you are trying to avoid.
- `eip`, `instance_container_restarts`, and `unhealthy_message` are already in the response and are the fields people reach for next.
- `DescribeApplicationInstances` accepts an `InstanceId` query param, so `ids` could be filtered server-side. Client-side is consistent with existing data sources, so this is genuinely optional.
- `Data.TotalSize` is returned for instances, so pagination could terminate on it rather than on `len(result) < PageSizeLarge` — the latter spins forever if the service ever stops honouring `CurrentPage`. Low risk, just sturdier.

xuzhang3 (changes requested):

PR #10156, opened 2026-08-07
resource/alicloud_instance: support instance_metadata_tags attribute

api-tool-agent:
## Description

Add a new top-level `instance_metadata_tags` attribute to `alicloud_instance` to expose the ECS instance tag feature in instance metadata, backed by the `InstanceMetadataTags` parameter of the ECS `ModifyInstanceMetadataOptions` API.

The field is flat at the top level (consistent with the existing `http_tokens` / `http_endpoint` / `http_put_response_hop_limit` attributes that already flatten the API `MetadataOptions` struct).

## Schema

- `instance_metadata_tags` (Optional, Computed): whether to expose the tags of the instance in the instance metadata.
- Valid values: `enabled`, `disabled`. Default value: `disabled`.
- Not `ForceNew` — the value can be changed in place.

## CRUD behavior

- **Create**: `RunInstances` does not support `InstanceMetadataTags`. The create path ends with `resourceAliCloudInstanceUpdate`, so once the instance reaches `Running` the existing `ModifyInstanceMetadataOptions` block applies the value — the same mechanism that already delivers `http_endpoint` / `http_tokens` on create. No separate create-time call is added.
- **Read**: `DescribeInstances` returns `InstanceMetadataTags` under `MetadataOptions` when `AdditionalAttributes` includes `META_OPTIONS` (already set by the existing `DescribeInstance` / `DescribeEcsInstance` helpers). The field is read from the raw response map because the SDK `MetadataOptions` struct does not yet expose it.
- **Update**: the existing `ModifyInstanceMetadataOptions` `HasChange` block (previously `http_endpoint` / `http_tokens`) is extended to also react to `instance_metadata_tags` and include `InstanceMetadataTags` in the request.

## Tests

Extended `TestAccAliCloudECSInstanceMetadataOptions` to set `instance_metadata_tags = "enabled"` on create and to toggle it back to `disabled` to exercise the update path.

## Docs

Added the attribute to `website/docs/r/instance.html.markdown` next to the other metadata-options attributes.


alicloud/resource_alicloud_instance.go
```diff
@@ -2356,6 +2372,7 @@ func resourceAliCloudInstanceUpdate(d *schema.ResourceData, meta interface{}) er
 
 		d.SetPartial("http_endpoint")
 		d.SetPartial("http_tokens")
+		d.SetPartial("instance_metadata_tags")
```
LCLSpring: 为什么要加这行？这是sdkv1的语法吧

PR #10172, opened 2026-08-10
resource/alicloud_oss_bucket_object: rebuild after out-of-band delete

api-tool-agent:
## Problem

When an OSS object is removed out-of-band (for example, via the OSS console or CLI) while it is still tracked by `alicloud_oss_bucket_object` in Terraform state, running `terraform plan` / `terraform refresh` aborts with an error instead of detecting the drift and planning a recreate.

## Root cause

`resourceAlicloudOssBucketObjectRead` matched the `404 Not Found` / `NoSuchKey` branch and called `d.SetId("")` (correctly asking Terraform to drop the resource from state), but then returned an error:

```go
if IsExpectedErrors(err, []string{"404 Not Found", "NoSuchKey"}) {
    d.SetId("")
    return WrapError(Error("To get the Object: ... not exist ...")) // aborts refresh
}
```

Under the Terraform plugin SDK, returning an error from `Read` discards the pending `SetId("")` state update and aborts the refresh/plan. The resource stays stuck in state and Terraform cannot recover without manual state removal.

## Fix

Return `nil` on the not-found branch so the SDK commits `SetId("")` during refresh. A subsequent plan then reports `1 to add` and `apply` recreates the object. Other errors still propagate as before.

## Acceptance test

Adds `TestAccAliCloudOssBucketObject_outOfBandDeleteRebuild`:

1. Create a bucket and object through Terraform.
2. Delete the object directly through the OSS API during the test check, simulating a console or CLI deletion.
3. Mark the step with `ExpectNonEmptyPlan: true`; the SDK's mandatory follow-up refresh must drop the object from state and plan a recreate. Before the fix, this refresh returns an error.
4. Re-apply the same configuration and verify the object is recreated.

The existing acceptance-test names are normalized to `TestAccAliCloud...`, and their configuration now covers the object header attributes required by the testing-coverage check. The documentation also clarifies that `object_worm_mode` is immutable because its only supported value is `COMPLIANCE`.

### DIFF
--- a/alicloud/resource_alicloud_oss_bucket_object.go
+++ b/alicloud/resource_alicloud_oss_bucket_object.go
@@ -285,8 +285,15 @@ func resourceAlicloudOssBucketObjectRead(d *schema.ResourceData, meta interface{
 	object, err := bucket.GetObjectDetailedMeta(d.Get("key").(string), options...)
 	if err != nil {
 		if IsExpectedErrors(err, []string{"404 Not Found", "NoSuchKey"}) {
+			// The object was removed out-of-band (e.g. via the OSS console or
+			// CLI) and no longer exists. SetId("") asks Terraform to drop it
+			// from state; return nil so the SDK commits that update during
+			// refresh. A subsequent plan then reports "1 to add" and apply
+			// recreates the object. Returning an error here would discard the
+			// SetId("") change and abort the refresh/plan, leaving the
+			// resource stuck in state and blocking further automation.
 			d.SetId("")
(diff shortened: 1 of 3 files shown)


ChenHanZhang (approved):

ChenHanZhang (approved):

api-tool-agent (commented):
approved

ChenHanZhang (approved):

ChenHanZhang (approved):

Issue #9961, opened 2026-07-10
[Major Version]: Terraform AliCloud Provider v2.0.0 (WIP)


xuzhang3:
# [Major Version]: Terraform AliCloud Provider v2.0.0

v2 stable will be released on October 12, 2026. After this release, v1 will enter maintenance mode — no new features will be added, and only critical security updates will be provided.

## Description

Version `2.0.0` of the Terraform AliCloud provider is a major release that includes breaking changes practitioners must account for before upgrading. This issue tracks every practitioner-facing change so the community can review the scope of the upgrade and provide feedback.

The first `2.0.0` release will be published as a **beta pre-release** (`2.0.0-beta1`). During the beta period we will keep the `1.x` line up to date with new service/resource coverage and bug fixes, gather community feedback, and resolve any issues found before promoting `2.0.0` to a stable release. After `2.0.0` reaches a stable release, the `1.x` line will receive critical fixes only.

The breaking changes in v2.0.0 are **narrow and uniform**: they are **schema type migrations from `TypeMap` to `TypeList`** on a set of resources and data sources. The upgrade requires HCL configuration edits and updates to any downstream references (outputs, `for`/`for_each` expressions, module wiring).

Upgrade instructions are in [Version 2 Upgrade Guide ](https://github.com/aliyun/terraform-provider-alicloud/blob/master/website/docs/guides/version-2-upgrade.html.markdown).

## Opting into the v2.0.0 beta

The beta build is versioned as a pre-release and uploaded to the Terraform Registry. Terraform only selects a pre-release when it is targeted by an **exact** constraint. Range or pessimistic operators (`~>`, `>=`, `>`) will **never** pick up a beta. To opt in, pin the pre-release version explicitly:

```terraform
terraform {
  required_providers {
    alicloud = {
      source  = "aliyun/alicloud"
      version = "= 2.0.0-beta1"
    }
  }
}
```

Notes and pitfalls:

- **All constraints must allow the beta.** Every constraint on the `alicloud` provider across the root module *and* any child modules must permit the pre-release. During beta testing, set every such constraint to `= 2.0.0-beta1` (or remove it).
- **Exact string match required.** The constraint must match the published tag exactly — `2.0.0-beta1` and `2.0.0-beta.1` are different pre-release identifiers.
- **Terraform version.** Pre-release constraints require a modern Terraform CLI; very old versions (e.g. 0.12) do not support them.

## Disclaimer — Test the Beta in Non-Production Only

[...]

PR #10198, opened 2026-08-12
feat: serve a terraform-plugin-framework provider alongside SDK v2

xuzhang3:
## What

Mux a `terraform-plugin-framework` provider with the existing `terraform-plugin-sdk/v2` one behind `tf5muxserver`, so new types can be authored framework-native without migrating the ~1130 SDK v2 resources and ~740 data sources. Protocol version, provider address and provider schema are unchanged, and `Provider()` itself is untouched.

This also makes the framework-only concept types — ephemeral resources, provider functions, list resources, actions — reachable for the first time.

## How

- **`alicloud/provider/framework`** — plumbing only: the framework provider, the mux factory, and a converter that derives the framework provider schema from the SDK v2 schema map at runtime. Deriving it is not a convenience: `tf5muxserver` refuses to serve providers whose provider schemas differ, and `endpointsSchema()` alone contributes 160 fields that would drift out of a hand-written mirror on the first new endpoint. Anything the converter has not been taught raises an error diagnostic rather than producing a silently divergent schema.
- **Server order is load-bearing.** The SDK v2 server is passed to `NewMuxServer` first, so its `ConfigureContextFunc` has built the `*connectivity.AliyunClient` by the time the framework `Configure` runs; the client is handed over through `(*schema.Provider).Meta()`. Credential resolution is not duplicated.
- **`alicloud/service/ims`** — the first framework-served type, `alicloud_ims_default_domain`. Framework types live in per-product service packages, never in the plumbing package: a service package may import `alicloud/connectivity` and nothing else of the provider, because `package alicloud`'s own tests reach `framework` and the reverse edge would not compile.
- **`alicloud/acctest`** — `ProtoV5ProviderFactories` and `PreCheck`: the exported test support a `<product>_test` package needs in order to reach a framework-served type.
- All six registration methods are declared, so registering the first ephemeral resource, list resource, action or provider function is a one-line change.

## Compatibility

No user-facing change. A plugin binary exposes type names over gRPC, not Go symbols — and here the provider schema is *derived* from the SDK v2 one and asserted equal by a unit test, the provider address and protocol version are unchanged, and no existing resource or data source is modified.

## Tests

Acceptance test for the new data source:

```
=== RUN   TestAccAliCloudImsDefaultDomainDataSource
--- PASS: TestAccAliCloudImsDefaultDomainDataSource (4.20s)
PASS
ok      github.com/aliyun/terraform-provider-alicloud/alicloud/service/ims      6.201s
```

**1 passed / 0 failed / 0 skipped.**

Unit tests added, covering the part most likely to break silently — the schema derivation:

- `TestUnitFrameworkProviderSchemaMatchesSDK` — the derived framework provider schema is served without `tf5muxserver` rejecting it
- `TestUnitProviderSchema` — per-field mapping from the SDK v2 schema
- `TestUnitProviderSchemaUnsupported` — an untaught construct raises a diagnostic instead of being dropped

### DIFF
--- a/CHANGELOG-v2.md
+++ b/CHANGELOG-v2.md
@@ -1,3 +1,10 @@
+## 2.0.0-beta3 (Unreleased)
+
+FEATURES:
+
+- **Terraform Plugin Framework:** the provider binary now serves a `terraform-plugin-framework` provider alongside the existing `terraform-plugin-sdk/v2` one, muxed behind `tf5muxserver`. Protocol version, provider address and provider schema are unchanged; new resources and data sources can be authored framework-native, and the framework-only concept types (ephemeral resources, provider functions, list resources, actions) become available for the first time.
+- **New Data Source:** `alicloud_ims_default_domain` — the default domain of the Alibaba Cloud account, backed by the Ims `GetDefaultDomain` API. Served by the framework provider.
+
 ## 2.0.0-beta2 (August 4, 2026)
 
 This beta rolls up every change merged from the 1.x line since v2.0.0-beta1 — see the `1.287.0` section of [CHANGELOG.md](CHANGELOG.md) — plus the v2-only fix below.
--- a/alicloud/acctest/acctest.go
+++ b/alicloud/acctest/acctest.go
@@ -1,0 +1,51 @@
+package acctest
+
+import (
+	"context"
+	"log"
+	"os"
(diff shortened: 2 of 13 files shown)


xuzhang3:
Add support for terraform-plugin-framework(#9961)

PR #10206, opened 2026-08-13
resource/alicloud_oss_bucket_cname: fix regression test; connectivity: stop discarding OSS API errors

ChenHanZhang:
## Summary

`TestAccAliCloudOssBucketCname_basic8544` failed with `InvalidDomainName.NoExist`,
because it bound `songwenpeng.alivetest.asia` — a zone hosted in a different
account. Correcting the domain uncovered three further defects, two of them in
the provider itself. All are fixed here.

The provider changes matter beyond this test:

- **`connectivity.Do` discarded every OSS API error** that came back alongside a
  non-nil response map. `normalizeOssOpenAPIResponse` only reshapes the map, but
  it was assigned to `err`, overwriting the error from the API call with its own
  `nil`. Failed OSS requests therefore became silent successes holding empty
  data. This one bug produced three different symptoms while debugging this test:
  a `PutCname` that "returned 200" without creating a binding, `unknown key
  CnameToken` when reading a token already consumed by `PutCname`, and `unknown
  key ListCnameResult` when verifying a cname was destroyed. Each of those has an
  explicit handler in the calling code (`NeedVerifyDomainOwnership`,
  `NoNeedCreateCnameToken`, `NoSuchBucket`) that could never fire.

- **`delete_certificate = true` could never succeed.** The update path sent
  `PreviousCertId`, `Force` and `DeleteCertificate` unconditionally, and OSS
  answers `MalformedXML` to any request holding both `DeleteCertificate` and
  `Force`, whatever the `Force` value. The failure was invisible while `Do` was
  discarding OSS errors, so the provider reported success and left the
  certificate in place.

## Test changes (basic8544)

- Bind a per-run unique subdomain of `aliterraform.com`, a zone that exists in the
  test account. OSS allows a domain to be bound to only one bucket globally, so a
  fixed domain would collide between concurrent runs and stay wedged after any run
  that failed before cleanup.
- Replace both certificates: the previous ones were issued for `tftestacc.com` and
  `bj.dinary.top` and expired in early 2025. The new ones are
  `*.aliterraform.com` wildcards, so any per-run subdomain is covered.
- Run in `ap-southeast-1` via `checkoutSupportedRegions`. Binding a custom domain
  to a bucket in a mainland China region requires the domain to have completed ICP
  filing, which no domain in the test account has; OSS reports this as
  `NoSuchCnameInRecord`, whose documented meaning is "域名未备案".
- Set `force = true` when rotating the certificate, as case 8386 already does.
  OSS rejects a replacement carrying neither `PreviousCertId` nor `Force` with
  `CasRenewCertificateConflict`.

## Validation (AccTest / ACube)

Target case:

| Resource | Case | taskId | Result |
|---|---|---|---|
| alicloud_oss_bucket_cname | TestAccAliCloudOssBucketCname_basic8544 | 9583 | passed (4/4 steps) |

Because `connectivity.Do` is shared, every resource that calls
`client.Do("Oss")` (29 files, 100 call sites) was swept: **27 resource codes,
60 cases — 58 passed, 1 skipped, 1 failed.**

Passed: AccessPoint, AccountPublicAccessBlock, BucketAccessMonitor,
BucketArchiveDirectRead, BucketCnameToken, BucketCors,
BucketDataRedundancyTransition, BucketHttpsConfig, BucketInventories,
BucketInventory, BucketLogging, BucketMetaQuery, BucketObjectWormConfiguration,
BucketOverwriteConfig, BucketPolicy, BucketPublicAccessBlock, BucketReferer,
BucketRequestPayment, BucketResponseHeader, BucketServerSideEncryption,
BucketStyle, BucketTransferAcceleration, BucketUserDefinedLogFields,
BucketVersioning, BucketWebsite, BucketWorm.

The single failure, `TestAccAliCloudOssBucketAcl_basic6192`, is **pre-existing
and not a regression**. Verified by reverting only `client.go` to master and
re-running the same case:

| Branch | taskId | Result |
|---|---|---|
| master's `client.go` | 9633 | failed — `Attribute 'acl' expected "public-read", got "private"` |
| this PR | 9605 | failed — `403 AccessDenied: Put public bucket acl is not allowed` |

Both fail. The underlying cause is an account/region restriction on public bucket
ACLs, not code. The comparison does show what this PR buys: the same failure went
from an unexplained attribute mismatch to the actual API rejection. The other
three `BucketAcl` cases pass.

## Test plan

- [x] Remote AccTest passes for `TestAccAliCloudOssBucketCname_basic8544` (taskId 9583)
- [x] Remote AccTest swept across all `client.Do("Oss")` resources (26/27 green)
- [x] Sole failure attributed to a pre-existing issue via a single-variable baseline
- [x] `make ci-check` — 5/6 pass; the Go Vet failure (`scripts/testing`: `main redeclared`) reproduces identically on untouched master, and `alicloud/connectivity` vets clean
- [ ] CI on this branch

## Not addressed

`TestAccAliCloudOssBucketCname_basic8542` and `_basic8386` share the same dead
domain and additionally reference `cert_id` values (`22495571-cn-hangzhou`,
`22495586-cn-hangzhou`) that no longer exist in the test account; 8386's inline
certificate also expired on 2026-03-28. Fixing them needs a certificate
provisioned in the dependency config rather than a hardcoded id, so they are left
for a follow-up.


ChenHanZhang:
Superseded by #10209, which carries the same change on a correctly-signed branch. Closing this one to avoid duplicate review.

PR #10208, opened 2026-08-13
resource/alicloud_route_entry: serialize create/delete and stop abandoned retry from creating after failure

ChenHanZhang:
## Summary

Fixes two related failure modes of `alicloud_route_entry` under concurrent creation (multiple route entries in one plan, e.g. VBR + RouterInterface topologies), both reproduced locally with `TF_LOG=DEBUG` evidence:

1. **Spurious final `TaskConflict` failure.** The VPC service processes route entry operations as serialized asynchronous tasks and rejects concurrent submissions with `TaskConflict` ("The operation is too frequent"). Concurrent sibling resources burn each other's retry budget on these rejections; when the retry window is short (e.g. the provider-level `max_retry_timeout` is set), the budget runs out while the server is still busy and the apply fails even though nothing is actually wrong.

2. **Hidden create after reported failure (root cause of `Duplicated.VpcNextHop` on the next apply).** `helper/resource.Retry` does not cancel an in-flight callback when its deadline expires. A callback blocked in `WaitForAllRouteEntriesAvailable` keeps running after Terraform has already reported the resource as failed, then sends `CreateRouteEntry` (same ClientToken) which can succeed on the cloud - leaving a route that exists server-side with no record in state. Reproduced: attempt 1 returned `TaskConflict` and the apply failed; the abandoned callback sent attempt 2 with the same ClientToken **64 seconds later** and received 200 OK; the route existed on the cloud while absent from state, so the next apply failed with `Duplicated.VpcNextHop`.

## Changes

- Add a process-wide mutex (`routeEntryTaskLock`) serializing `CreateRouteEntry`/`DeleteRouteEntry` calls. Queueing happens before the retry loop starts, so waiting for a sibling operation does not consume the resource's own retry budget. `TaskConflict` stays in the retryable lists because other actors (a second Terraform run, the console, other tools) can still trigger it.
- Add an `abandoned` guard (atomic flag set right after `resource.Retry` returns, checked in the callback between `WaitForAllRouteEntriesAvailable` and `CreateRouteEntry`), so a callback that outlives the retry deadline can no longer create the route after the outcome has been decided.

## Verification

- Repro before fix (v1.288.0, template with several route entries, `max_retry_timeout = 10`, `-parallelism=10`): apply failed with final `TaskConflict` on one route entry; TF_LOG shows the abandoned attempt 2 succeeding 64s after the failure was reported, route present on cloud but not in state.
- Same template with this patch (dev_overrides build): `Apply complete! Resources: 24 added, 0 changed, 0 destroyed.`; all 5 route entries created on attempt 1 with zero `TaskConflict` responses; all present in state.
- `go build ./alicloud/...`, `go vet`, `gofmt` clean; existing unit tests pass:
  `go test ./alicloud -run 'TestUnit(IsRouteEntryWaitTimeout|BuildRouteEntryResourceId)$'`

### DIFF
--- a/alicloud/resource_alicloud_route_entry.go
+++ b/alicloud/resource_alicloud_route_entry.go
@@ -3,6 +3,8 @@ package alicloud
 import (
 	"fmt"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/vpc"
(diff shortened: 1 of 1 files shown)


ChenHanZhang (commented):
approved

ChenHanZhang (commented):
approved

PR #10243, opened 2026-08-18
New Resource: alicloud_event_bridge_event_streaming

api-tool-agent:
## Summary

- Add a new resource `alicloud_event_bridge_event_streaming` to manage EventBridge event streamings, which continuously stream events from an event source to an event sink, with optional event filtering, transformation and runtime options.
- `source`, `filter_pattern`, `sink`, `run_options` and `transforms` are JSON strings that follow the EventBridge API definitions, mirroring how `alicloud_event_bridge_rule` models `filter_pattern`.
- The `status` attribute drives the stream lifecycle through `StartEventStreaming` / `PauseEventStreaming` (`RUNNING` / `PAUSED`), and tags are managed through `TagResources` / `UntagResources`. Import is supported.

## Test plan

- [x] Local `go vet` on the changed package and `terraform validate` of the documentation example against the locally built provider both pass.
- [x] Remote acceptance tests:

```
=== RUN TestAccAliCloudEventBridgeEventStreaming_basic0
--- PASS: TestAccAliCloudEventBridgeEventStreaming_basic0 (141.26s)

=== RUN TestAccAliCloudEventBridgeEventStreaming_basic0_twin
--- PASS: TestAccAliCloudEventBridgeEventStreaming_basic0_twin (20.72s)
```

`TestAccAliCloudEventBridgeEventStreaming_basic0`: create -> update description/source/filter_pattern/sink/run_options/transforms -> status RUNNING -> status PAUSED -> tags create/update/remove -> import -> destroy.
`TestAccAliCloudEventBridgeEventStreaming_basic0_twin`: create with all attributes including status RUNNING and tags -> import -> destroy.


alicloud/resource_alicloud_event_bridge_event_streaming.go
```diff
@@ -0,0 +1,367 @@
...
+		Schema: map[string]*schema.Schema{
+			"event_streaming_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"source": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateFunc:     validateEventBridgeEventStreamingJson,
```
xuzhang3: validjson 可以用官方json校验工具

alicloud/resource_alicloud_event_bridge_event_streaming.go
```diff
@@ -0,0 +1,367 @@
...
+	return resourceAliCloudEventBridgeEventStreamingRead(d, meta)
+}
+
+func resourceAliCloudEventBridgeEventStreamingRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*connectivity.AliyunClient)
+	eventBridgeService := EventBridgeServiceV2{client}
+
+	object, err := eventBridgeService.DescribeEventBridgeEventStreaming(d.Id())
+	if err != nil {
+		if !d.IsNewResource() && NotFoundError(err) {
```
xuzhang3: read 不需要  d.IsNewResource()

alicloud/resource_alicloud_event_bridge_event_streaming.go
xuzhang3, line 39: Please use the stock validator `validation.StringIsJSON` here and at the other four `ValidateFunc` usages (`filter_pattern`, `sink`, `run_options`, `transforms`), and delete `validateEventBridgeEventStreamingJson` - the hand-rolled version accepts the empty string where the stock form rejects it. The `DiffSuppressFunc` has no stock equivalent and can stay.

alicloud/resource_alicloud_event_bridge_event_streaming.go
```diff
@@ -0,0 +1,367 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"status": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: StringInSlice([]string{"RUNNING", "PAUSED"}, false),
+			},
+			"tags": tagsSchema(),
```
xuzhang3, line 78: Please use `tagsSchemaString()` instead - new code should use the Elem-carrying form (`TypeMap` with `Elem: TypeString`); `tagsSchema()` is TypeMap without Elem and its own definition carries `// lintignore: S006`.

alicloud/resource_alicloud_event_bridge_event_streaming.go
```diff
@@ -0,0 +1,367 @@
...
+	update := false
+
+	request := map[string]interface{}{
+		"EventStreamingName": d.Id(),
+	}
+
+	if d.HasChange("description") {
+		update = true
+	}
+	if v, ok := d.GetOk("description"); ok {
```
xuzhang3, line 222: `description` is the only update parameter gated by `GetOk` while every other parameter is always sent - on removal, Description is omitted from the request. If the API keeps the value when the parameter is absent, removing `description` from the config becomes a permanent diff. Please always send it (empty string on removal) and add a removal step to basic0; this changes request wiring, so the ACC run needs a refresh after the fix.

alicloud/resource_alicloud_event_bridge_event_streaming.go
```diff
@@ -0,0 +1,367 @@
...
+		request["Description"] = v
+	}
+
+	// Source, FilterPattern and Sink are required by UpdateEventStreaming, so
```
xuzhang3, line 226: The comment says Source/FilterPattern/Sink are required by `UpdateEventStreaming`, but the official contract marks them optional (only `EventStreamingName` is required). The always-send behavior is the right direction - only the comment needs updating.

website/docs/r/event_bridge_event_streaming.html.markdown
```diff
@@ -0,0 +1,106 @@
...
+  })
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `event_streaming_name` - (Required, ForceNew) The name of the event streaming.
+* `source` - (Required, Json) The event provider, which is also known as the event source. You must and can specify only one event source. The value is a JSON string, for example `{"SourceMNSParameters":{"QueueName":"example","RegionId":"cn-hangzhou","IsBase64Decode":true}}`.
```
xuzhang3, line 76: The schema type is `TypeString`, so the label should be `String`, not `Json` - the JSON-ness is already described in prose below. Same for `filter_pattern`, `sink`, `run_options`, `transforms`.

website/docs/r/event_bridge_event_streaming.html.markdown
```diff
@@ -0,0 +1,106 @@
...
+
+# alicloud_event_bridge_event_streaming
+
+Provides a Event Bridge Event Streaming resource.
+
+An event streaming allows you to stream events from an event source to an event target, with optional event filtering and transformation.
+
+For information about Event Bridge Event Streaming and how to use it, see [What is Event Streaming](https://www.alibabacloud.com/help/en/eventbridge/latest/api-eventbridge-2020-04-01-createeventstreaming).
+
+-> **NOTE:** Available since v1.290.0.
```
xuzhang3, line 18: `v1.290.0` is stale - the unreleased CHANGELOG top is 1.292.0. Please bump to `v1.292.0`.

xuzhang3 (changes requested):
Request changes. One blocker: the `description` update path (inline) - it is the only update parameter gated by `GetOk` while every other parameter always-sends, so removing `description` from the config omits Description from the request; the removal transition is also untested. Please always send it, add a removal step to basic0, and refresh the ACC run after the fix (this one changes request wiring). The other fixes inline are self-contained and don't touch request wiring, so the existing evidence (PASS: basic0 141.26s, basic0_twin 20.72s) stays valid for those.

PR #10259, opened 2026-08-19
resource/alicloud_mongodb_sharding_audit_filter: New Resource: alicloud_mongodb_sharding_audit_filter, New Data Source: alicloud_mongodb_sharding_audit_filters

api-tool-agent:
## What

Adds a new resource `alicloud_mongodb_sharding_audit_filter` and the companion data source `alicloud_mongodb_sharding_audit_filters` for managing the audit log filter of an ApsaraDB for MongoDB **sharding cluster** instance.

The resource is a singleton configuration bound to a sharding cluster instance (`db_instance_id`). It exposes the audit log filter (`filter`), the audit policy state (`audit_status`, `service_type`, `storage_period`, `hot_storage_period`) and the sharding-specific `role_type`, and imports by `db_instance_id`.

## Schema

| field | type | required | notes |
|---|---|---|---|
| `db_instance_id` | string | yes, ForceNew | sharding cluster instance id; also the import key |
| `audit_status` | string | yes | `enable` / `disabled` |
| `filter` | string | yes | comma-separated; `admin`, `slow`, `query`, `insert`, `update`, `delete`, `command` |
| `role_type` | string | no, Computed | `mongos` (mongos node) / `db` (shard node) |
| `storage_period` | int | no, Computed | audit log retention in days |
| `hot_storage_period` | int | no, Computed | hot storage duration in days (V2_Standard only) |
| `service_type` | string | no, Computed | `Standard`, `V2_Standard`; ignored while `audit_status = disabled` (server flips edition on disable) |
| `region_id` | string | Computed | region of the sharding cluster instance |

## CRUD mapping

Singleton configuration resource (no real Create/Delete on the API side):

- **Create / Update** — `ModifyAuditPolicy` (audit_status / storage_period / hot_storage_period / service_type) and `ModifyAuditLogFilter` (filter / role_type).
- **Read** — `DescribeMongoDBLogConfig` + `DescribeAuditPolicy` + `DescribeAuditLogFilter`.
- **Delete** — no-op; Terraform removes the resource from state, the server-side config remains.
- **Import** — `terraform import alicloud_mongodb_sharding_audit_filter.example <db_instance_id>`.

Mirrors the existing `alicloud_mongodb_audit_policy` (replica-set) implementation; the only delta is the `role_type` field and that the backing instance is a sharding cluster.

## Notes

- `role_type` is Optional+Computed; the Provider only forwards a non-empty value to `ModifyAuditLogFilter` so the Optional+Computed drift (empty config vs API-read state) does not push an unintended value.
- `DescribeAuditLogFilter` returns a merged view (`mongos@admin,slow-db@admin,slow`) with `RoleType=logic` when `RoleType` is omitted on read. Read parses out the segment matching the configured `role_type`, and falls back to the value common to all roles when `role_type` is unset (e.g. on import), so state matches the config instead of the raw merged string.
- `ModifyAuditPolicy` requires `StoragePeriod` and `HotStoragePeriod` to be sent together. Both are forwarded unconditionally on Update so that changing only one of them (or re-enabling audit with neither set in the config) does not produce a partial request.
- `service_type` keeps the same DiffSuppressFunc + disabled-preserve Read branch as `alicloud_mongodb_audit_policy`, since the same server-side quirk (edition flipped while audit is off) applies.
- While the audit log is disabled, `DescribeMongoDBLogConfig` omits the retention fields; Read keeps the state value instead of writing 0 over it, which would otherwise cause a perpetual diff.

## Tests

Acceptance tests (remote execution):

```
=== RUN TestAccAliCloudMongodbShardingAuditFilter_basic0
--- PASS: TestAccAliCloudMongodbShardingAuditFilter_basic0 (2868.97s)

=== RUN TestAccAliCloudMongodbShardingAuditFilter_basic0_twin
--- PASS: TestAccAliCloudMongodbShardingAuditFilter_basic0_twin (1576.43s)

=== RUN TestAccAliCloudMongodbShardingAuditFiltersDataSource
--- PASS: TestAccAliCloudMongodbShardingAuditFiltersDataSource (1650.63s)
```

- `TestAccAliCloudMongodbShardingAuditFilter_basic0` — create with all fields, update filter / storage_period / hot_storage_period, disable and re-enable, import.
- `TestAccAliCloudMongodbShardingAuditFilter_basic0_twin` — create with `role_type` set, import.
- `TestAccAliCloudMongodbShardingAuditFiltersDataSource` — data source read backed by the resource.
- `TestUnitAliCloudMongodbShardingAuditFilter` — unit test covering Create/Update/Read/Delete mock paths; runs in CI.


website/docs/r/mongodb_sharding_audit_filter.html.markdown
```diff
@@ -0,0 +1,112 @@
...
+  Provides a Alicloud Mongodb Sharding Audit Filter resource.
+---
+
+# alicloud_mongodb_sharding_audit_filter
+
+Provides a Mongodb Sharding Audit Filter resource.
+
+For information about Mongodb Sharding Audit Filter and how to use it, see [ModifyAuditLogFilter](https://www.alibabacloud.com/help/en/mongodb/developer-reference/api-mongodb-2015-12-01-modifyauditlogfilter).
+
+-> **NOTE:** Available since v1.234.0.
```
LCLSpring: 这个是新资源，这里的版本号不对吧

api-tool-agent: Good catch — fixed in 592e592de.

Both docs said `Available since v1.234.0`, which is long released. Updated the resource and the data source to `v1.290.0`, the current unreleased version (`1.289.0` shipped on August 17, 2026). This matches the convention of the recently added resources, which annotate the unreleased version at the time the PR is opened.

PR #10265, opened 2026-08-20
resource/alicloud_fcv3_custom_domain: support binding HTTPS certificate by id

api-tool-agent:
## What does this PR do?

Adds an optional `certificate_id` attribute to `alicloud_fcv3_custom_domain`. When set, the provider resolves the certificate and private key from the referenced SSL certificate managed by SSL Certificates Service (CAS) and binds them as the HTTPS certificate for the custom domain. This lets users reference a managed certificate without pasting the PEM material.

- `certificate_id` is `(Optional, Computed)` and conflicts with `cert_config.0.certificate` and `cert_config.0.private_key`.
- The resolved private key is never persisted to state; `cert_config` is populated from the Function Compute response.

## Test cases

```
=== RUN TestAccAliCloudFcv3CustomDomain_certificateId_basic
--- PASS: TestAccAliCloudFcv3CustomDomain_certificateId_basic (9.36s)
```

The acceptance test `TestAccAliCloudFcv3CustomDomain_certificateId_basic` covers the `certificate_id` create and rotation (Create + Update resolution) paths, plus import. The test uploads two self-signed certificates to SSL Certificates Service, creates the custom domain referencing the first by id, rotates to the second, and imports the resource.

### DIFF
--- a/alicloud/resource_alicloud_fcv3_custom_domain.go
+++ b/alicloud/resource_alicloud_fcv3_custom_domain.go
@@ -79,6 +79,12 @@ func resourceAliCloudFcv3CustomDomain() *schema.Resource {
 					},
 				},
 			},
+			"certificate_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"cert_config.0.certificate", "cert_config.0.private_key"},
+			},
 			"cors_config": {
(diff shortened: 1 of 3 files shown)


ChenHanZhang (approved):

ChenHanZhang (approved):

shanye997:
image

PR #10271, opened 2026-08-20
New Resource: alicloud_polardb_gateway

kittenshow98:


alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+		Delete: resourceAlicloudPolarDBGatewayDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(50 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"region_id": {
```
xuzhang3, line 25: region_id 可以从 client 直接获取，这里是有自定义需求吗？ 如果 region 和 client 的 region 不一致，endpoint不对，服务端走不通的吧

alicloud/resource_alicloud_polardb_gateway_test.go
```diff
@@ -0,0 +1,117 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					rac.resourceAttrMapUpdateSet()(map[string]string{
+						"region_id": CHECKSET, "zone_id": CHECKSET, "db_cluster_class": "polar.mysql.x4.large",
+						"pay_type": "Postpaid", "auto_renew": "false", "vpc_id": CHECKSET,
+						"period": "Month", "used_time": "1",
+						"vswitch_id": CHECKSET, "security_group_id": CHECKSET, "db_type": "MySQL",
+						"status": CHECKSET, "description": CHECKSET, "create_time": CHECKSET,
+						"modify_time": CHECKSET, "expire_time": CHECKSET, "expired": CHECKSET,
+						"latest_version": CHECKSET, "current_version": CHECKSET, "running_version": CHECKSET,
+						"endpoints": CHECKSET, "security_ip_arrays": CHECKSET,
```
xuzhang3: **These two assertions can never pass.**

`CHECKSET` maps to `TestCheckResourceAttrSet`, which does a literal flatmap lookup — sdk v1 `helper/resource/testing.go`:

```go
func testCheckResourceAttrSet(is *terraform.InstanceState, name string, key string) error {
	if val, ok := is.Attributes[key]; !ok || val == "" {
		return fmt.Errorf("%s: Attribute '%s' expected to be set", name, key)
	}
```

There is no `.#` sugar. For a `TypeList` the state only ever contains `endpoints.#`, `endpoints.0.address`, … — the bare key `endpoints` is never present, so both checks fail unconditionally.

House style for list assertions is `"endpoints.#": "1"` plus `"endpoints.0.address": CHECKSET`; there is no other place in the repo that asserts a bare list name.

xuzhang3 (commented):
Read the files at head, plus the published `polardb` 2017-08-01 definitions for `CreateGateway`, `DescribeGatewayAttribute` and `DeleteGateway`, and compared against `polardb_aicluster` as the nearest template.

The overall shape holds up: with no `Modify*` API in the gateway surface (Create / Delete / DescribeAttribute / DescribeList / ApikeyList / RestartPolarClawGateway), having no Update and marking every argument `ForceNew` is the right call — worth a one-line comment saying so, so nobody later "fixes" it. `Status` values `CREATE`/`ACTIVATION` check out against the API doc, and `polarDBGatewayList` correctly tolerates both the wrapped and the plain-array response shapes.

Comments are inline. The two that matter most:

- The acceptance test cannot pass as written — six assertions are provably wrong, independent of the service's behaviour. Please run it and put the results in the PR description.
- `region_id` as a `Required` argument diverges from every other resource in this provider, and schema cannot be walked back after release.

Smaller things I did not comment on individually: `SecurityIPArrayTag` is dropped from the `security_ip_arrays` mapping; the import example ID prefix is `gw-` where the API documents `pg-`; `port` is declared `TypeInt` while the API returns a string; the unit test fixture only feeds the wrapped response shape and a numeric port, so it exercises the branch that probably never fires; `CheckDestroy` resolves the region differently from the resource; and the one-line-per-multiple-fields schema formatting diverges from the rest of the repo.

Verification gap: I could not enumerate valid gateway class codes or the real not-found error code from the published definitions — the two comments resting on those say so.

alicloud/resource_alicloud_polardb_gateway_test.go
```diff
@@ -0,0 +1,117 @@
...
+				}),
+				Check: resource.ComposeTestCheckFunc(
+					rac.resourceAttrMapUpdateSet()(map[string]string{
+						"region_id": CHECKSET, "zone_id": CHECKSET, "db_cluster_class": "polar.mysql.x4.large",
+						"pay_type": "Postpaid", "auto_renew": "false", "vpc_id": CHECKSET,
+						"period": "Month", "used_time": "1",
+						"vswitch_id": CHECKSET, "security_group_id": CHECKSET, "db_type": "MySQL",
+						"status": CHECKSET, "description": CHECKSET, "create_time": CHECKSET,
+						"modify_time": CHECKSET, "expire_time": CHECKSET, "expired": CHECKSET,
+						"latest_version": CHECKSET, "current_version": CHECKSET, "running_version": CHECKSET,
```
xuzhang3: **Four more assertions contradict the published API response.**

With `pay_type = "Postpaid"` on line 67:

- `expire_time` — `DescribeGatewayAttribute` documents `ExpireTime` as empty when the billing method is Postpaid. `CHECKSET` fails on `""`.
- `latest_version` / `current_version` / `running_version` — all three come back as empty strings in the API's own response demo.
- `description` maps from `GwDescription`, and the resource exposes no way to set a description at create time.

Together with #1 and the empty PR body, this reads as a case that has not been executed yet. Please run it and paste the `PASS: TestAcc… (12.34s)` lines plus the pass/fail tally into the PR description — that output would also settle the two API-shape questions below.

alicloud/resource_alicloud_polardb_gateway_test.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+		IDRefreshName: resourceId,
+		Providers:     testAccProviders,
+		CheckDestroy:  rac.checkResourceDestroy(),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccConfig(map[string]interface{}{
+					"region_id":         "${data.alicloud_regions.default.regions.0.id}",
+					"zone_id":           "${data.alicloud_vswitches.default.vswitches.0.zone_id}",
+					"db_cluster_class":  "polar.mysql.x4.large",
```
xuzhang3: **`polar.mysql.x4.large` looks like a PolarDB *cluster* class, not a gateway class.**

Both `CreateGateway.DBClusterClass` and `DescribeGatewayAttribute.ClassCode` document `polar.app.g2.medium` as their example value. If a cluster class is rejected here, step 0 fails at create and the case can never go green.

The same value is the only example in the new doc page (`polardb_gateway.html.markdown`), so if it is wrong it misleads every user who copies it.

Disclosure: there is no `DescribeGatewayClasses`-style API, so I cannot enumerate the valid values — please confirm against a successful create.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+				Type: schema.TypeString, Required: true, ForceNew: true,
+				Description: "The region ID of the PolarDB gateway.",
+			},
```
xuzhang3, line 28: **`region_id` should not be a required argument.**

Every other resource in this provider declares `region_id` as `Computed` only and takes the region from the provider block — `resource_alicloud_alb_load_balancer.go`, `resource_alicloud_cen_transit_router.go`, `resource_alicloud_click_house_enterprise_db_cluster.go`, and so on. Schema is permanent: once a new resource ships with a `Required` argument, it cannot be removed without a breaking change.

Three concrete problems:

1. Users must restate the provider's region on every resource. When the two disagree, `RpcPost` still resolves the endpoint from the *provider's* region, so the gateway is created somewhere other than what state records.
2. It is `ForceNew` **and** written back by Read (see the separate note on the Read mapping) — the classic replacement footgun.
3. After `terraform import`, `d.Get("region_id")` is empty, so `DescribePolarDBGatewayAttributeWithRegion` omits `RegionId` and falls back to the client region. That path works — which is itself evidence the argument is not needed.

Suggested shape: make `region_id` `Computed` only, and use `client.RegionId` in Create/Delete, exactly as the new `DescribePolarDBGatewayAttribute` wrapper already does.

`polardb_aicluster` and `polardb_application` took the same shortcut, but neither of their Read functions depends on the value in state; this one does.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+		"status": "Status", "description": "GwDescription", "create_time": "CreateTime",
+		"modify_time": "ModifyTime", "expire_time": "ExpireTime", "expired": "Expired",
+		"latest_version": "LatestVersion", "current_version": "CurrentVersion", "running_version": "RunningVersion",
+	} {
+		if value, ok := object[responseKey]; ok && value != nil {
+			if err = d.Set(key, value); err != nil {
+				return WrapError(err)
+			}
+		}
+	}
```
xuzhang3, line 212: **`ForceNew` + Read write-back = silent destroy/recreate.**

This mapping writes `ClassCode` → `db_cluster_class`, `PayType` → `pay_type`, `DbType` → `db_type`, `VpcId` → `vpc_id`, `VSwitchId` → `vswitch_id` and `RegionId` → `region_id` back into state — and all six of those fields are declared `ForceNew`.

The failure chain: apply → the API normalises the value → Read writes the actual value into state → the next plan diffs config against state → any difference in representation is a **replacement** of a billed gateway, with no user-visible cause.

`db_cluster_class` is the highest risk (the API returns a normalised class code), `pay_type` next. Either make the written-back arguments `Optional + Computed` instead of `ForceNew`, or stop writing them back.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+			},
+			"db_cluster_class": {
+				Type: schema.TypeString, Optional: true, ForceNew: true,
+				Description: "The specifications of the PolarDB gateway.",
+			},
+			"pay_type": {
+				Type: schema.TypeString, Required: true, ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{"Postpaid", "Prepaid"}, false),
+				Description:  "The billing method. Valid values: `Postpaid`, `Prepaid`.",
+			},
```
xuzhang3, line 41: **Prefer the repo's own validators over `helper/validation`.**

`validators.go` defines `StringInSlice`, `IntBetween`, `IntAtLeast` and `StringMatch` for a specific reason: they check `skipResourceSchemaValidation()`, so `TF_SKIP_RESOURCE_SCHEMA_VALIDATION=true` downgrades the error to a warning. That is the only escape hatch a user has when the provider's hardcoded enum lags behind the service — and PolarDB adding a third `pay_type` or a third `db_type` is exactly when it is needed. `helper/validation` has no such hatch.

`resource_alicloud_polardb_aicluster.go` — the file this resource is modelled on — uses the bare `StringInSlice`. Applies to `pay_type`, `period`, `db_type` and the `IntBetween` on `used_time`, and lets you drop the `helper/validation` import.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+		"AutoRenew": d.Get("auto_renew"),
+	}
+	for key, requestKey := range map[string]string{
+		"zone_id": "ZoneId", "db_cluster_class": "DBClusterClass", "period": "Period",
+		"used_time": "UsedTime", "security_group_id": "SecurityGroupId", "db_type": "DBType",
+	} {
+		if value, ok := d.GetOk(key); ok {
+			request[requestKey] = value
+		}
+	}
```
xuzhang3, line 161: **`Period` and `UsedTime` are sent even when the billing method is Postpaid.**

The loop forwards whatever is set, without consulting `pay_type`. The API documents both parameters as meaningful only for `Prepaid`, and nothing rejects the reverse combination (`Postpaid` + `period`), so a config that mixes them applies cleanly and silently does nothing.

Either gate them on `pay_type == "Prepaid"` when building the request, or add `ConflictsWith` so it fails at plan time. Note the acceptance test currently sets `period`/`used_time` together with `pay_type = "Postpaid"`, which means the Prepaid required-argument check above is never exercised.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+		response, err = client.RpcPost("polardb", "2017-08-01", action, nil, request, false)
+		if err != nil {
+			if NeedRetry(err) {
+				wait()
+				return resource.RetryableError(err)
+			}
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
```
xuzhang3: **The create retry has no idempotency token.**

`resource.Retry` + `NeedRetry` means a request that reached the service but whose response was lost gets sent again — producing a second, billed gateway that Terraform does not know about.

Disclosure: the published `CreateGateway` parameter list has no `ClientToken`, so the fix here is narrowing the retryable error set rather than adding a token. Worth confirming with the service team whether `ClientToken` is supported but undocumented — `polardb_aicluster` does pass one to `CreateAIDBCluster`.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	stateConf := BuildStateConf([]string{"CREATE", "ACTIVATION", "DELETING"}, []string{}, d.Timeout(schema.TimeoutDelete), 10*time.Second, service.PolarDBGatewayStateRefreshFunc(d.Id(), d.Get("region_id").(string), []string{}))
+	if _, err = stateConf.WaitForState(); err != nil {
+		if strings.HasPrefix(err.Error(), "couldn't find resource") {
+			d.SetId("")
+			return nil
+		}
+		return WrapErrorf(err, IdMsg, d.Id())
+	}
```
xuzhang3: **The delete wait signals success through an SDK-internal error string, and burns ~20 polls doing it.**

`PolarDBGatewayStateRefreshFunc` returns `nil, "", nil` on not-found, and `Target` is empty — so this lands in the `notfoundTick` branch of `StateChangeConf.WaitForState`. `NotFoundChecks` defaults to 20 and `BuildStateConf` sets `MinTimeout: 3s`, so after the gateway is already gone the wait keeps polling for another ~21 rounds before returning a `*NotFoundError`, which is then recognised by prefix-matching `"couldn't find resource"`. Every destroy pays about a minute for nothing, and the happy path is an error path.

Cleaner: have the refresh function return a sentinel on not-found (`return "DELETED", "DELETED", nil`) and pass `Target: []string{"DELETED"}` — the wait then completes on the first poll with no error and no string matching.

This idiom currently exists only in `polardb_aicluster` and `polardb_application`, both added recently, so it is not an established repo convention.

alicloud/service_alicloud_polardb.go
```diff
@@ -2297,6 +2297,60 @@ func (s *PolarDBService) DescribePolarDBAIClusterAttribute(id string) (object ma
 	return object, nil
 }
 
+var polarDBGatewayNotFoundCodes = []string{"InvalidGwClusterId.NotFound", "InvalidGatewayId.NotFound", "ResourceNotFound", "Gateway.NotFound"}
```
xuzhang3: Two things on this line.

**Placement.** `service_alicloud_polardb.go` has no top-level `var` or `const` today (78 functions, zero declarations). This is the first one, and it lands after roughly forty functions. Package-level `const`/`var` belong above every `func`, immediately after the imports — please move it to the top of the file. No pre-existing debt to inherit here.

**Content.** These four codes look like a guess. If none of them matches what the service actually returns, the consequences compound rather than degrade:

- `NotFoundError(err)` never fires in Read, so a gateway deleted out of band makes `terraform plan` error instead of planning a recreate;
- the `IsExpectedErrors` idempotency branch in Delete never fires;
- `PolarDBGatewayStateRefreshFunc` never reports not-found, so the delete wait cannot complete and runs to the 10-minute timeout.

Please confirm the code against a real 404 response.

alicloud/resource_alicloud_polardb_gateway_test.go
```diff
@@ -0,0 +1,117 @@
...
+variable "name" {
+  default = "%s"
+}
+
+data "alicloud_regions" "default" {
+  current = true
+}
+
+data "alicloud_vswitches" "default" {
+}
```
xuzhang3: **Unfiltered data source — flaky by construction.**

`data "alicloud_vswitches" "default" {}` with no arguments takes whatever vSwitch happens to come back first, in an arbitrary zone, which may not be able to host a gateway at all. The other dependence configs in the polardb test files either create their own VPC/vSwitch or filter with `alicloud_zones`.

That also undercuts the comment above: the point of reusing an existing vSwitch is to save VPC quota, but an arbitrary one costs you a reliable test instead.

alicloud/resource_alicloud_polardb_gateway.go
```diff
@@ -0,0 +1,293 @@
...
+				Type: schema.TypeString, Computed: true,
+				Description: "The running gateway version.",
+			},
+			"endpoints": {
+				Type: schema.TypeList, Computed: true,
+				Elem: &schema.Resource{Schema: map[string]*schema.Schema{
+					"address":      {Type: schema.TypeString, Computed: true},
+					"endpoint_id":  {Type: schema.TypeString, Computed: true},
+					"gateway_id":   {Type: schema.TypeString, Computed: true},
+					"port":         {Type: schema.TypeInt, Computed: true},
```
xuzhang3: `port` should be `TypeString` instead of `TypeInt`. The SDK uses `json.NewDecoder(...).UseNumber()` when parsing API responses, which returns `json.Number` values. The `d.Set` → `mapstructure` path rejects `json.Number` for `TypeInt` fields, causing `Read` to fail on every refresh. Sibling resources like `alicloud_mongodb_sharding_network_public_address` also use `TypeString` for port fields.

kittenshow98:
image

PR #10277, opened 2026-08-20
resource/alicloud_realtime_compute_vvp_instance: support auto_renew for subscription instance

api-tool-agent:
## Summary
- Add `renew_status`, `auto_renew_duration`, and `renewal_duration_unit` arguments to `alicloud_realtime_compute_vvp_instance` to manage subscription instance renewal through the BssOpenApi `SetRenewal` action.
- `renew_status` accepts `AutoRenewal`, `ManualRenewal`, `NotRenewal`; `auto_renew_duration` accepts `1`, `2`, `3`, `6`, `12` (validated via `IntInSlice`); `renewal_duration_unit` accepts `M` (month) and `Y` (year). These arguments only take effect when `payment_type = "Subscription"`.
- Renewal state is read back from BssOpenApi `QueryAvailableInstances` (`RenewStatus`, `RenewalDuration`, `RenewalDurationUnit`) because the foasconsole `DescribeInstances` API does not return renewal fields.
- Domestic and international accounts are dispatched via `IsInternationalAccount` with the appropriate bss `ProductType`.
- Replaced deprecated `GetOkExists` with `GetOk` + `> 0` guard for `auto_renew_duration` to avoid sending `RenewalPeriod = 0` when the field is unset.

## Test plan
- [x] `TestAccAliCloudRealtimeComputeVvpInstance_basic4594` — Subscription with renewal trio (create + `auto_renew_duration` update); PASS (remote ACC, 833.02s).
- [ ] `TestAccAliCloudRealtimeComputeVvpInstance_basic4594_intl` — International-site subscription with renewal trio; guarded by `testAccPreCheckWithAccountSiteType(t, IntlSite)`. Skipped at the IntlSite gate during remote ACC because the acceptance account is domestic (cn-hangzhou) and lacks international-site subscription capability — not a code defect; pending international-site account access enablement (does not block domestic merge).
- [ ] `TestAccAliCloudRealtimeComputeVvpInstance_basic4636` — PayAsYouGo regression check.

### DIFF
--- a/alicloud/resource_alicloud_realtime_compute_vvp_instance.go
+++ b/alicloud/resource_alicloud_realtime_compute_vvp_instance.go
@@ -12,6 +12,25 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
+// bss (BssOpenApi 2017-12-14) renewal ProductCode/ProductType for the VVP
+// (Realtime Compute for Apache Flink) subscription instance.
+//
+// Domestic values empirically confirmed via bss GetOrderDetail on two ACC VVP
+// orders (ProductCode=sc, ProductType=sc_flinkserverless_public_cn,
+// CommodityCode=sc_flinkserverless_public_cn, SubscriptionType=Subscription,
+// PaymentStatus=Paid) and bss QueryAvailableInstances on a live instance
+// (returned RenewStatus=ManualRenewal, RenewalDurationUnit=M).
+//
+// The intl ProductType follows the domestic _cn → _intl naming convention used
+// by other alicloud subscription resources (e.g. elasticsearchpre →
+// elasticsearchpre_intl). Confirmed by the Flink product owner 2026-09-03;
+// international-site ACC verification is still pending (see PR test plan).
+const (
+	realtimeComputeVvpBssProductCode     = "sc"
+	realtimeComputeVvpBssProductType     = "sc_flinkserverless_public_cn"
+	realtimeComputeVvpBssProductTypeIntl = "sc_flinkserverless_public_intl" // confirmed by the Flink product owner 2026-09-03
+)
+
 func resourceAliCloudRealtimeComputeVvpInstance() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceAliCloudRealtimeComputeVvpInstanceCreate,
@@ -27,6 +46,11 @@ func resourceAliCloudRealtimeComputeVvpInstance() *schema.Resource {
 			Delete: schema.DefaultTimeout(35 * time.Minute),
 		},
 		Schema: map[string]*schema.Schema{
+			"auto_renew_duration": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				ValidateFunc: IntInSlice([]int{1, 2, 3, 6, 12}),
+			},
 			"create_time": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -45,6 +69,16 @@ func resourceAliCloudRealtimeComputeVvpInstance() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				ValidateFunc: StringInSlice([]string{"Month"}, false),
+			},
+			"renew_status": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"renewal_duration_unit": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
 			},
 			"resource_group_id": {
 				Type:     schema.TypeString,
@@ -234,6 +268,10 @@ func resourceAliCloudRealtimeComputeVvpInstanceCreate(d *schema.ResourceData, me
 		return WrapErrorf(err, IdMsg, d.Id())
 	}
 
+	if err := setRealtimeComputeVvpInstanceRenewal(d, meta); err != nil {
+		return WrapError(err)
+	}
+
 	return resourceAliCloudRealtimeComputeVvpInstanceRead(d, meta)
 }
 
@@ -300,6 +338,21 @@ func resourceAliCloudRealtimeComputeVvpInstanceRead(d *schema.ResourceData, meta
 	}
 
 	d.Set("vswitch_ids", vSwitchIds1Raw)
+
+	if checkValue := d.Get("payment_type"); checkValue == "Subscription" {
+		// The foasconsole DescribeInstances API does not return renewal state, so
+		// renewal is read back from bss QueryAvailableInstances. bss uses a
+		// different naming from SetRenewal: RenewStatus / RenewalDuration /
+		// RenewalDurationUnit (vs. RenewalStatus / RenewalPeriod / RenewalPeriodUnit).
+		bssOpenApiService := BssOpenApiService{client}
+		renewRaw, err := bssOpenApiService.QueryAvailableInstances(d.Id(), client.RegionId, realtimeComputeVvpBssProductCode, realtimeComputeVvpBssProductType, realtimeComputeVvpBssProductCode, realtimeComputeVvpBssProductTypeIntl)
+		if err != nil && !NotFoundError(err) {
+			return WrapError(err)
+		}
+		d.Set("renew_status", renewRaw["RenewStatus"])
+		d.Set("auto_renew_duration", renewRaw["RenewalDuration"])
+		d.Set("renewal_duration_unit", renewRaw["RenewalDurationUnit"])
+	}
 
 	return nil
 }
@@ -405,6 +458,9 @@ func resourceAliCloudRealtimeComputeVvpInstanceUpdate(d *schema.ResourceData, me
 		}
 		d.SetPartial("tags")
 	}
+	if err := setRealtimeComputeVvpInstanceRenewal(d, meta); err != nil {
+		return WrapError(err)
+	}
 	d.Partial(false)
 	return resourceAliCloudRealtimeComputeVvpInstanceRead(d, meta)
 }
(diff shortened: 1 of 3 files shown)


alicloud/resource_alicloud_realtime_compute_vvp_instance.go
```diff
@@ -12,6 +12,26 @@ import (
...
+//
+// The intl ProductType follows the domestic _cn → _intl naming convention used
+// by other alicloud subscription resources (e.g. elasticsearchpre →
+// elasticsearchpre_intl). It is an inference, not empirically verified against
+// an international-site instance — confirm with international-site ACC or the
+// product-side billing config before relying on it.
+const (
+	realtimeComputeVvpBssProductCode     = "sc"
+	realtimeComputeVvpBssProductType     = "sc_flinkserverless_public_cn"
+	realtimeComputeVvpBssProductTypeIntl = "sc_flinkserverless_public_intl"
```
xuzhang3: This constant is an inference (as the comment above it documents) — but it is load-bearing for this feature: the originating requirement is from an international-site customer, so `SetRenewal` on the target accounts uses exactly this value, and a wrong ProductType fails for precisely the customers this PR serves.

Before merge, please confirm `sc_flinkserverless_public_intl` against the intl billing config, or validate it with an intl-site ACC run. The domestic counterpart was verified empirically; the intl one hasn't.

alicloud/resource_alicloud_realtime_compute_vvp_instance.go
```diff
@@ -27,6 +47,10 @@ func resourceAliCloudRealtimeComputeVvpInstance() *schema.Resource {
 			Delete: schema.DefaultTimeout(35 * time.Minute),
 		},
 		Schema: map[string]*schema.Schema{
+			"auto_renew_duration": {
```
xuzhang3, line 49: The docs list `Valid values: 1, 2, 3, 6, 12` for this argument, but the schema has no `ValidateFunc`, so nothing enforces the range. Concretely: an explicit `auto_renew_duration = 0` passes the `GetOkExists` gate in `setRealtimeComputeVvpInstanceRenewal` and gets sent to `SetRenewal` as `RenewalPeriod = 0`.

Suggest `ValidateFunc: validation.IntInSlice([]int{1, 2, 3, 6, 12})` so invalid values are rejected at plan time instead of relying on the API.

alicloud/resource_alicloud_realtime_compute_vvp_instance_test.go
```diff
@@ -210,10 +210,13 @@ func TestAccAliCloudRealtimeComputeVvpInstance_basic4594(t *testing.T) {
...
 						"${data.alicloud_vswitches.default.ids.0}"},
-					"zone_id":       "cn-hangzhou-i",
-					"payment_type":  "Subscription",
-					"pricing_cycle": "Month",
-					"duration":      "1",
+					"zone_id":               "cn-hangzhou-i",
+					"payment_type":          "Subscription",
+					"pricing_cycle":         "Month",
+					"duration":              "1",
+					"renew_status":          "AutoRenewal",
```
xuzhang3, line 217: This case covers the domestic path (`ProductType = sc_flinkserverless_public_cn`). The intl path — `IsInternationalAccount() → sc_flinkserverless_public_intl` — currently has no coverage mechanism: all four cases pin the `cn-hangzhou` PreCheck and hardcode `zone_id = "cn-hangzhou-i"`.

Since the requirement comes from an international-site customer, could you add an intl-region variant of this case (renewal trio on create + `auto_renew_duration` change on update, mirroring this test)? Note there's no `realtime_compute_vvp_zones` data source yet, so the intl region's zone needs a fixed value.

xuzhang3 (commented):
Thanks — the renewal wiring itself checks out (create/update gating in `setRealtimeComputeVvpInstanceRenewal`, the bss read-back with RenewStatus/RenewalDuration/RenewalDurationUnit mapping, and test coverage of the positive / not-configured / PayAsYouGo paths). Two blockers before merge:

**1. PR description doesn't match the implementation.** The Summary describes an `auto_renew` boolean (Optional, ForceNew, mapping to the foasconsole `CreateInstance` `AutoRenew` parameter, added to `ImportStateVerifyIgnore`) — none of that exists in this PR. The shipped design is the bss `SetRenewal` trio (`renew_status` / `auto_renew_duration` / `renewal_duration_unit`) with read-back via `QueryAvailableInstances`. Even the Test plan references `auto_renew = true`, which isn't an attribute in the code. Please rewrite the Summary and Test plan to match the actual implementation.

**2. ACC evidence is still "pending".** Both Test plan checkboxes are unchecked (`remote ACC pending`). A domestic ACC run of `TestAccAliCloudRealtimeComputeVvpInstance_basic4594` on this branch passed (~658s) — please check the boxes and record the results as `PASS: <test name> (<duration>)`. The intl-side run is the remaining gap: the requirement originates from an international-site customer, so that run is what actually validates the target scenario (see the inline comments on the intl ProductType constant and test coverage).

alicloud/resource_alicloud_realtime_compute_vvp_instance.go
```diff
@@ -454,6 +510,75 @@ func resourceAliCloudRealtimeComputeVvpInstanceDelete(d *schema.ResourceData, me
...
+	client := meta.(*connectivity.AliyunClient)
+	if v, ok := d.GetOk("payment_type"); !ok || v.(string) != "Subscription" {
+		return nil
+	}
+	needUpdate := false
+	if d.IsNewResource() {
+		if _, ok := d.GetOk("renew_status"); ok {
+			needUpdate = true
+		}
+		if _, ok := d.GetOkExists("auto_renew_duration"); ok {
```
xuzhang3: Follow-up to the schema comment: since the suggested `ValidateFunc` will reject `0` at plan time, this call can simply become `GetOk` with a `> 0` guard:

```go
if v, ok := d.GetOk("auto_renew_duration"); ok && v.(int) > 0 {
	needUpdate = true
}
```

`GetOkExists` is marked deprecated in the SDK — both the v1 and v2 lines carry `Deprecated: usage is discouraged due to undefined behaviors and may be removed in a future version of the SDK` — so new code shouldn't add new call sites. The `> 0` guard provides the same configured-detection with deterministic semantics.

alicloud/resource_alicloud_realtime_compute_vvp_instance.go
```diff
@@ -454,6 +510,75 @@ func resourceAliCloudRealtimeComputeVvpInstanceDelete(d *schema.ResourceData, me
...
+	if client.IsInternationalAccount() {
+		request["ProductType"] = realtimeComputeVvpBssProductTypeIntl
+	}
+	if v, ok := d.GetOk("auto_renew_duration"); ok && v.(int) > 0 {
+		request["RenewalPeriod"] = v
+	}
+	if v, ok := d.GetOk("renewal_duration_unit"); ok {
+		request["RenewalPeriodUnit"] = v
+	}
+	if v, ok := d.GetOk("renew_status"); ok {
```
xuzhang3, line 552: When only `renew_status = "AutoRenewal"` is configured (create with the trio incomplete, or an update changing it alone), the request goes out with `RenewalStatus` but no `RenewalPeriod` / `RenewalPeriodUnit` — the SetRenewal contract marks both as required when `RenewalStatus=AutoRenewal`, so the failure surfaces at the API instead of locally. The bastionhost and elasticsearch SetRenewal call sites both guard this locally (error out when AutoRenewal is requested without the companion params). Suggest the same shape here: before building the request, if the effective `renew_status` is `AutoRenewal` and either companion wouldn't ride along, return a local error naming the missing argument.

website/docs/r/realtime_compute_vvp_instance.html.markdown
```diff
@@ -91,9 +91,12 @@ You can resume managing the subscription instance via the AlibabaCloud Console.
 ## Argument Reference
 
 The following arguments are supported:
+* `auto_renew_duration` - (Optional) The auto-renewal period of the subscription instance. It only takes effect when `payment_type = "Subscription"` and `renew_status = "AutoRenewal"`. Valid values: `1`, `2`, `3`, `6`, `12`. The unit is specified by `renewal_duration_unit`.
 * `duration` - (Optional) The number of subscription periods. If the payment type is PRE, this parameter is required.
 * `payment_type` - (Required, ForceNew) The payment type of the resource.
 * `pricing_cycle` - (Optional) The subscription period. If the payment type is PRE, this parameter is required.
+* `renew_status` - (Optional, Computed) The renewal status of the subscription instance. It only takes effect when `payment_type = "Subscription"`. Valid values: `AutoRenewal`, `ManualRenewal`, `NotRenewal`.
```
xuzhang3, line 98: The `Valid values` claim here and on `renewal_duration_unit` (`M`/`Y`) isn't enforced by the schema — neither attribute has a `ValidateFunc` (only `auto_renew_duration` carries `IntInSlice`). Suggest `StringInSlice([]string{"AutoRenewal", "ManualRenewal", "NotRenewal"}, false)` and `StringInSlice([]string{"M", "Y"}, false)` — validation only binds user config, so the read-back path is unaffected. (Alternatively, drop the docs claims.)

website/docs/r/realtime_compute_vvp_instance.html.markdown
```diff
@@ -91,9 +91,12 @@ You can resume managing the subscription instance via the AlibabaCloud Console.
 ## Argument Reference
 
 The following arguments are supported:
+* `auto_renew_duration` - (Optional) The auto-renewal period of the subscription instance. It only takes effect when `payment_type = "Subscription"` and `renew_status = "AutoRenewal"`. Valid values: `1`, `2`, `3`, `6`, `12`. The unit is specified by `renewal_duration_unit`.
```
xuzhang3, line 94: The three new entries lack `Available since` markers (house style on this file). The current unreleased version is v1.293.0 — please align with whichever version is unreleased at merge time.

xuzhang3 (commented):
The renewal wiring follows the shipped design (bss `SetRenewal` trio with `QueryAvailableInstances` read-back), with `TestAccAliCloudRealtimeComputeVvpInstance_basic4594` recorded as PASS (833.02s) and the intl variant's account-gate skip documented. No blockers — the remaining comments are non-blocking suggestions.

PR #10295, opened 2026-08-26
resource/alicloud_pvtz_user_vpc_authorization: fix nil auth_type in resource ID

api-tool-agent:
## Problem

When `auth_type` is omitted, the resource ID was built using `request["AuthType"]`, which returns nil for the unset map key. `fmt.Sprint(nil)` produces the literal string `<nil>`, so the state ID becomes `<authorized_user_id>:<nil>`. Subsequent Read and Delete operations parse this ID and send `<nil>` as `AuthType` to the `DescribeUserVpcAuthorizations` and `DeleteUserVpcAuthorization` APIs, causing failures.

## Fix

- **Create**: Use `d.Get("auth_type")` instead of `request["AuthType"]` to build the ID, defaulting to `NORMAL` when the optional field is omitted. This matches the `DeleteUserVpcAuthorization` API default.
- **Read/Delete**: Migrate legacy state IDs containing `<nil>` to `NORMAL` so existing resources continue to work without manual state edits.
- **Docs**: Note that `auth_type` defaults to `NORMAL`.

## Test

Added unit test `CreateNormalWithDefaultAuthType` that creates the resource without setting `auth_type` and asserts the ID is `<authorized_user_id>:NORMAL` rather than `<authorized_user_id>:<nil>`.

### DIFF
--- a/alicloud/resource_alicloud_pvtz_user_vpc_authorization.go
+++ b/alicloud/resource_alicloud_pvtz_user_vpc_authorization.go
@@ -33,6 +33,7 @@ func resourceAlicloudPvtzUserVpcAuthorization() *schema.Resource {
 			"auth_type": {
 				Type:         schema.TypeString,
 				Optional:     true,
+				Computed:     true,
 				ForceNew:     true,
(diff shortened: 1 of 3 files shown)


shanye997:
image

Issue #2131, opened 2020-01-21
Imported instance resources will trigger change when running apply


liuyangc3:
### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

```
terraform -v
Terraform v0.12.18
+ provider.alicloud v1.67.0
+ provider.external v1.2.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- alicloud_instance

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl

```

### Debug Output
no
### Panic Output
no
### Expected Behavior
What should have happened?

apply should return no errors

### Actual Behavior
What actually happened?

The instance will be to change but will got error after type `yes`.
```
Plan: 1 to add, 1 to change, 0 to destroy.

 you want to perform these actions in workspace "defualt"?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

Error: [ERROR] terraform-provider-alicloud/alicloud/resource_alicloud_instance.go:637:
[ERROR] terraform-provider-alicloud/alicloud/resource_alicloud_instance.go:1219: Resource i-hp37ousm8e8c5zmpiuho ModifyInstanceAttribute Failed!!! [SDK alibaba-cloud-sdk-go ERROR]:
SDK.ServerError
ErrorCode: Credit.NotFound
Recommend:
RequestId: E55171AE-C220-43B1-BFAF-8AD2538F5533
Message: The specified credit information does not exist.
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
0. create an instance with security_enhancement_strategy = "Active", save the instance id.
1. create resource alicloud_instance.foo in main.tf , and import it `terraform import alicloud_instance.foo i-<instance_id>`
2. add any new resources
3 `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

the imported instance attribute 
`security_enhancement_strategy` is null in the state file.

cofyc:
@xiaozhu36 This is an annoying issue. `alicloud_instance` is one of the core resources. This makes this provider unusable in many cases.

PR #10304, opened 2026-08-27
resource/alicloud_instance: suppress spurious security_enhancement_strategy ForceNew diff after import

api-tool-agent:
## Summary

`alicloud_instance.security_enhancement_strategy` is a create-only field: it is accepted by
RunInstances/CreateInstance/ReplaceSystemDisk, but `DescribeInstance` does not return it, so the
resource `Read` never writes it back to state.

After `terraform import`, the state value is always empty. When the config still declares the
field (e.g. `security_enhancement_strategy = "Active"`), the plan shows a spurious ForceNew diff
that needlessly recreates the instance on every apply.

## Fix

Extend the existing `resourceAliCloudInstanceCustomizeDiff` hook to clear the
`security_enhancement_strategy` diff for any existing instance (non-create pass) when the state
value is empty and the config value is non-empty, treating the config value as the source of
truth the way import does. The create-time behavior is unchanged (the field is still sent on
RunInstances); only the post-import spurious diff is suppressed.

## Test

A new acceptance test `TestAccAliCloudECSInstanceSesImportNoop` is added:

1. Create an instance with `security_enhancement_strategy = "Active"`.
2. Import it (the field is ignored in ImportStateVerify since it is not returned by
   DescribeInstance).
3. Re-apply the same config; the step expects an empty plan. Without the hook this step would
   recreate the instance (ForceNew diff); with the hook it is a no-op.

### DIFF
--- a/alicloud/resource_alicloud_instance.go
+++ b/alicloud/resource_alicloud_instance.go
@@ -2717,10 +2717,23 @@ func resourceAliCloudInstanceDelete(d *schema.ResourceData, meta interface{}) er
 	return nil
 }
 
-// resourceAliCloudInstanceCustomizeDiff turns an image_id change into a replacement when the
-// provider is configured with features.ecs_instance.replace_on_image_update. Without it, such a
-// change is applied in place by modifyInstanceImage, which the plan can only describe as an update.
+// resourceAliCloudInstanceCustomizeDiff does two jobs:
+//   - Suppress the spurious ForceNew diff on security_enhancement_strategy after import. The
+//     field is create-only (RunInstances/CreateInstance/ReplaceSystemDisk accept it, but
+//     DescribeInstance does not return it), so Read never writes it back; after import the state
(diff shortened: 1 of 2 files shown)


api-tool-agent:
Closing in favor of #10307, which fixes the same spurious `security_enhancement_strategy` ForceNew diff after import and additionally covers the `credit_specification` half of the problem (the `Credit.NotFound` error from `ModifyInstanceAttribute` on non-burstable instances reported in #2131). #10307 uses field-level `DiffSuppressFunc` (consistent with existing provider precedents such as `ecsSystemDiskPerformanceLevelSuppressFunc`), ships documentation notes for both attributes, and has remote acceptance test evidence (`TestAccAliCloudECSInstanceImportedAttributeDiff` PASS, `TestAccAliCloudECSInstanceVpc` PASS). Thank you for the earlier attempt — the `CustomizeDiff` approach here only addressed the SES field.

PR #10312, opened 2026-08-27
resource/alicloud_oss_bucket_object: accept hex content_md5 and convert to base64

api-tool-agent:
## Summary
- `alicloud_oss_bucket_object.content_md5` now accepts a hex-encoded MD5 digest (as produced by Terraform's `md5()` / `filemd5()` functions) and converts it to the base64 form required by the OSS PutObject `Content-MD5` header (RFC 1864). Previously the value was forwarded verbatim, so a hex digest made `PutObject` fail with a digest mismatch and the field was effectively unusable with `md5()`/`filemd5()`.
- Values already base64-encoded are passed through unchanged after validating that the decoded value is a 16-byte MD5 digest; a 32-character hex MD5 digest is converted to base64. Any other input is rejected at the provider side with a clear error instead of being forwarded to OSS.
- The new acceptance test for hex `content_md5` reuses the existing test fixture and references `alicloud_oss_bucket.default.bucket` directly, consistent with the current object test dependence.

## Test plan
- [x] `TestUnitNormalizeContentMD5` (unit): hex->base64 conversion, base64/16-byte validation, and invalid input rejection. PASS: `TestUnitNormalizeContentMD5` (≈6s, taskId=10223).
- [x] `TestAccAliCloudOssBucketObject_contentMD5Hex` (ACC): creates an OSS bucket object with a hex `content_md5` and asserts `PutObject` succeeds. PASS: `TestAccAliCloudOssBucketObject_contentMD5Hex` (≈10s, taskId=10225).
- [x] `TestAccAliCloudOssBucketObject_basic` (ACC, existing): regression - base64 `content_md5` still accepted. PASS on Step0/Step1; Step2 (`acl=public-read`) fails with 403 when the test account has account-level Block Public Access enabled. This is a pre-existing environment constraint unrelated to `content_md5`.


Winphies (approved):
Reviewed the latest changes. The implementation now converts 32-character hex MD5 to base64, validates base64 input as a 16-byte MD5 digest, and includes corresponding unit/ACC coverage. LGTM.

ChenHanZhang (approved):

website/docs/r/oss_bucket_object.html.markdown
```diff
@@ -93,7 +93,7 @@ The following arguments are supported:
...
 * `content_encoding` - (Optional) Specifies what content encodings have been applied to the object and thus what decoding mechanisms must be applied to obtain the media-type referenced by the Content-Type header field. Read [RFC2616 Content-Encoding](https://www.ietf.org/rfc/rfc2616.txt) for further details.
-* `content_md5` - (Optional) The MD5 value of the content. Read [MD5](https://www.alibabacloud.com/help/doc-detail/31978.htm) for computing method.
+* `content_md5` - (Optional) The MD5 digest of the content for OSS data-integrity verification. Accepts a 32-character **hex** MD5 digest (as produced by Terraform's [`md5()`](https://developer.hashicorp.com/terraform/language/functions/md5) and [`filemd5()`](https://developer.hashicorp.com/terraform/language/functions/filemd5) functions; the provider converts it to base64), or a **base64**-encoded MD5 digest (RFC 1864, as produced by [`filebase64md5()`](https://developer.hashicorp.com/terraform/language/functions/filebase64md5)). Values that are neither a valid hex MD5 digest nor valid base64 are rejected by the provider. Read [MD5](https://www.alibabacloud.com/help/doc-detail/31978.htm) for the computing method.
```
xuzhang3: The `filebase64md5()` reference here doesn't resolve — Terraform has no base64-output MD5 function (the base64 forms exist only for SHA-256/SHA-512: `base64sha256`, `filebase64sha256`, `base64sha512`, `filebase64sha512`; `md5()` and `filemd5()` are hex-only), and the link 404s. Since hex is the Terraform-native form, suggest dropping the function citation, e.g.:

* `content_md5` - (Optional) The MD5 digest of the content for OSS data-integrity verification. Accepts a 32-character **hex** MD5 digest (as produced by Terraform's `md5()` / `filemd5()`; the provider converts it to the base64 form required by RFC 1864) or a **base64**-encoded MD5 digest. Values in neither form are rejected by the provider. Read [MD5](https://www.alibabacloud.com/help/doc-detail/31978.htm) for the computing method.

xuzhang3 (commented):
Thanks — the core fix is correct and nicely engineered: the hex branch ordered before base64 is provably unambiguous (a base64-encoded 16-byte digest always ends with `=` padding, which is not a hex character, so genuine base64 values can never be misparsed as hex); provider-side validation errors surface before any PUT in both the upload and Read paths; and values OSS accepted before this change are exactly the ones the passthrough branch accepts, so existing state upgrades cleanly. The unit vectors check out and the ACC fixture digest matches `md5()` of its content.

Two blockers, both about the PR description rather than the code:

**1. The description doesn't match the diff.**
- The third Summary bullet claims the `alicloud_oss_bucket_public_access_block` dependency was dropped from the object acceptance tests. The test-file diff is pure additions (+85/−0): the dependence function still creates the block resource, the existing cases still reference `${alicloud_oss_bucket_public_access_block.default.bucket}` (L51/L121/L269), and the new case creates the block via the same dependence. Either land that cleanup or drop the claim.
- The same bullet says invalid input "is left for OSS to reject" — the code rejects it provider-side in `normalizeContentMD5` (the unit test and the docs line below both say so).
- The Test plan names `TestNormalizeContentMD5`; the function is `TestUnitNormalizeContentMD5`.

**2. ACC evidence is pending.** Both Test plan checkboxes are unchecked — please run `TestAccAliCloudOssBucketObject_contentMD5Hex` and the `basic` regression and fill them in with `PASS: <test name> (<duration>)`.

Non-blocking: `content_md5` validation fires at apply time (consistent with this file's existing `expires` handling); a schema `ValidateFunc` mirroring the acceptance check would surface typos at plan time, matching `acl`/`server_side_encryption`/`object_worm_mode` — optional polish.

xuzhang3 (changes requested):

PR #10349, opened 2026-09-01
resource/alicloud_ecs_invocation: mark timed parameter as deprecated and update docs

api-tool-agent:
## Summary

The ECS `InvokeCommand` API has deprecated the `Timed` parameter — the API ignores its value regardless of what is passed. This change marks the `timed` schema field as `Deprecated` to emit a plan-time warning, and updates the resource documentation to reflect the latest API semantics.

## Changes

- **`alicloud/resource_alicloud_ecs_invocation.go`**: Added `Deprecated` attribute to the `timed` schema field. The field type (`TypeBool`), `Optional`, `ForceNew`, and `Computed` are unchanged. The field is retained for backward compatibility.
- **`website/docs/r/ecs_invocation.html.markdown`**: Updated the `repeat_mode`, `timed`, and `frequency` argument descriptions:
  - `repeat_mode` defaults to `Once` when `frequency` is not specified, and `Period` when `frequency` is specified.
  - `timed` is marked as **Deprecated** with migration guidance.
  - `frequency` description updated to recommend `repeat_mode=Period` for recurring execution.

## Backward Compatibility

This is a non-breaking change:
- The `timed` field remains in the schema — existing configurations continue to work with a deprecation warning.
- The `Deprecated` attribute only adds a plan-time warning; Create/Read behavior is unchanged.
- The `Create` function still passes `Timed` to the API (ignored by the API, same as before).
- The `Read` function still mirrors the `DescribeInvocations` response.

## Migration

Users who set `timed = true` with a `frequency` should switch to `repeat_mode = "Period"` with `frequency`.

### DIFF
--- a/alicloud/resource_alicloud_ecs_invocation.go
+++ b/alicloud/resource_alicloud_ecs_invocation.go
@@ -44,10 +44,11 @@ func resourceAlicloudEcsInvocation() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"Once", "Period", "NextRebootOnly", "EveryReboot"}, false),
 			},
 			"timed": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				ForceNew: true,
-				Computed: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				ForceNew:   true,
+				Computed:   true,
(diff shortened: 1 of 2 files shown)


api-tool-agent:
Closing as duplicate of #10353, which contains the same schema and documentation changes plus the acceptance-test fixes verified remotely. Please follow #10353 for this change.

PR #10358, opened 2026-09-01
resource/alicloud_mse_gateway: support resource_group_id and tags

### LINKED ISSUE #8757: mse gateway 不支持资源组与 tags
https://api.aliyun.com/api/mse/2019-05-31/AddGateway

希望可以在创建 MSE Gateway 的时候支持设置资源组和标签

api-tool-agent:
### Description

Add `resource_group_id` and `tags` support for the `alicloud_mse_gateway` resource.

- `resource_group_id` - (Optional, Computed) The ID of the resource group which the gateway belongs.
- `tags` - (Optional, Computed) A mapping of tags to assign to the resource.

The MSE AddGateway API already accepts `ResourceGroupId` and `Tag` parameters. The gateway Read path now populates both fields from the GetGateway response, and Update uses the generic ChangeResourceGroup and TagResources/UnTagResources APIs to keep them in sync.

### Related Issue
Fixes #8757

### DIFF
--- a/alicloud/resource_alicloud_mse_gateway.go
+++ b/alicloud/resource_alicloud_mse_gateway.go
@@ -1,6 +1,7 @@
 package alicloud
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"time"
@@ -72,6 +73,12 @@ func resourceAlicloudMseGateway() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"resource_group_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"tags": tagsSchema(),
 			"delete_slb": {
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -147,6 +154,13 @@ func resourceAlicloudMseGatewayCreate(d *schema.ResourceData, meta interface{}) 
 
 	request["VSwitchId"] = d.Get("vswitch_id")
 	request["Vpc"] = d.Get("vpc_id")
+	if v, ok := d.GetOk("resource_group_id"); ok {
+		request["ResourceGroupId"] = v
+	}
+	if v, ok := d.GetOk("tags"); ok {
+		tagsMap := ConvertTags(v.(map[string]interface{}))
+		request = expandTagsToMap(request, tagsMap)
+	}
 
 	wait := incrementalWait(3*time.Second, 3*time.Second)
 	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
@@ -196,6 +210,17 @@ func resourceAlicloudMseGatewayRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("status", fmt.Sprint(formatInt(object["Status"])))
 	d.Set("vswitch_id", object["Vswitch"])
 	d.Set("vpc_id", object["Vpc"])
+	d.Set("resource_group_id", object["ResourceGroupId"])
+	// GetGateway returns the gateway tags as a JSON string in the "MseTag" field
+	// (e.g. {"For":"Test","acs:rm:rgId":"rg-xxx","Created":"TF"}); there is no
+	// "Tags" array on this API. tagIgnored filters the acs: system tags.
+	var tagMap map[string]interface{}
+	if mseTagStr, ok := object["MseTag"].(string); ok && mseTagStr != "" {
+		if err := json.Unmarshal([]byte(mseTagStr), &tagMap); err != nil {
+			log.Printf("[DEBUG] Resource alicloud_mse_gateway unmarshal MseTag failed: %s", err)
+		}
+	}
+	d.Set("tags", tagsToMap(tagMap))
 
 	slbListObject, err := mseService.ListGatewaySlb(d.Id())
 	if err != nil {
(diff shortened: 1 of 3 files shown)


alicloud/resource_alicloud_mse_gateway.go
```diff
@@ -230,36 +255,81 @@ func resourceAlicloudMseGatewayRead(d *schema.ResourceData, meta interface{}) er
 }
 func resourceAlicloudMseGatewayUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*connectivity.AliyunClient)
-	request := map[string]interface{}{
-		"GatewayUniqueId": d.Id(),
+	mseService := MseService{client}
+	d.Partial(true)
```
xuzhang3, line 259: Please remove the new partial-state apparatus: `d.Partial(true)` here, the three `d.SetPartial(...)` calls (lines 290, 322, 329), and `d.Partial(false)` (line 332). New `SetPartial` usage is not accepted on the v1 SDK line — the SDK v2 conversion strips every call (#10346) — and with the window properly closed and Read persisting the full state at the end of Update, removal is behavior-neutral on the success path. The pre-existing `d.SetPartial("tags")` inside `SetResourceTags` is stock and stays.

alicloud/resource_alicloud_mse_gateway_test.go
```diff
@@ -135,42 +135,58 @@ func TestAccAlicloudMSEGateway_basic0(t *testing.T) {
...
-						"replica":      "2",
-						"vpc_id":       CHECKSET,
-						"vswitch_id":   CHECKSET,
-						"gateway_name": name,
+						"spec":              "MSE_GTW_2_4_200_c",
+						"replica":           "2",
+						"vpc_id":            CHECKSET,
+						"vswitch_id":        CHECKSET,
+						"gateway_name":      name,
+						"resource_group_id": CHECKSET,
```
xuzhang3, line 156: A set-only check cannot catch the accepted-but-ignored case: if AddGateway accepted but ignored `ResourceGroupId`, state would hold the default resource group id — still non-empty, test green, feature silently no-op. Please assert the configured value instead, e.g. `TestCheckResourceAttrPair("alicloud_mse_gateway.default", "resource_group_id", "data.alicloud_resource_manager_resource_groups.default", "groups.1.id")`.

alicloud/resource_alicloud_mse_gateway_test.go
```diff
@@ -135,42 +135,58 @@ func TestAccAlicloudMSEGateway_basic0(t *testing.T) {
...
+						"tags.For":     "Test-update",
 					}),
 				),
 			},
 			{
 				ResourceName:            resourceId,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"enterprise_security_group", "slb_spec", "delete_slb"},
+				ImportStateVerifyIgnore: []string{"enterprise_security_group", "slb_spec", "delete_slb", "resource_group_id"},
```
xuzhang3, line 184: `resource_group_id` is read back in Read (`d.Set("resource_group_id", ...)` from GetGateway), so imported state should match applied state. Please remove this entry and keep the import verify passing — if it fails after removal, that is a Read gap to fix, not a difference to hide. The other three entries are create-only parameters Read never sets and can stay.

alicloud/resource_alicloud_mse_gateway_test.go
```diff
@@ -282,13 +305,15 @@ func TestUnitAlicloudMSEGateway(t *testing.T) {
...
-			"Status":   "2",
+			"Spec":            "spec",
+			"Replica":         2,
+			"Vpc":             "vpc_id",
+			"Vswitch":         "vswitch_id",
+			"Vswitch2":        "backup_vswitch_id",
+			"Name":            "gateway_name",
+			"Status":          "2",
+			"ResourceGroupId": "resource_group_id",
+			"Tags":            map[string]interface{}{"Created": "TF", "For": "Test"},
```
xuzhang3, line 316: The mock response sets a `Tags` map, but the code reads `MseTag` — a JSON string field in the GetGateway response — so this feeds a key the implementation never reads, and the new JSON-parsing + system-tag-filtering path gets zero coverage. Please feed `MseTag` here as a JSON string (e.g. `{"Created":"TF","For":"Test","acs:rm:rgId":"rg-xxx"}`) and assert after create that `tags` equals `{"Created": "TF", "For": "Test"}` — proving both the parse and the `acs:` filtering.

website/docs/r/mse_gateway.html.markdown
```diff
@@ -73,6 +73,8 @@ The following arguments are supported:
 * `vswitch_id` - (Required, ForceNew) The ID of the vswitch.
 * `vpc_id` - (Required, ForceNew) The ID of the vpc.
 * `delete_slb` - (Optional) Whether to delete the SLB purchased on behalf of the gateway at the same time.
+* `resource_group_id` - (Optional, Computed) The ID of the resource group which the gateway belongs.
```
xuzhang3, line 76: Please add the version marker to both new attribute lines matching the current CHANGELOG unreleased version (`Available since v1.292.0` as of today — re-check when amending, releases land daily), and add `Map` to the tags line (see the `mse_cluster` tags entry for the form).

xuzhang3 (changes requested):
This PR adds `resource_group_id` and `tags` support to `alicloud_mse_gateway`. Requesting changes for missing evidence and coverage:

1. **Acceptance test evidence missing** — the body has no `Acceptance tests` section. Please run `TestAccAliCloudMSEGateway*` on the current head and add `PASS: <test name> (<duration>)` for each case.
2. **New `Partial`/`SetPartial` apparatus** — please remove the 5 new lines (`d.Partial(true)`, three `d.SetPartial(...)`, `d.Partial(false)`); the SDK v2 conversion strips them (#10346) and removal is behavior-neutral here.
3. **`ChangeResourceGroup` path untested** — no test step changes `resource_group_id` after create, so the new update block never executes. Please add a step that creates without `resource_group_id`, then sets it to a second resource group.
4. **`resource_group_id` in `ImportStateVerifyIgnore`** — it is read back in Read; please remove the entry and keep import verify passing.
5. **`CHECKSET` for `resource_group_id`** — please replace with a paired assertion against the data source so an accepted-but-ignored value fails the test.
6. **Unit-test mock feeds a dead key** — the mock sets `Tags` but the code parses `MseTag` (JSON string); please feed `MseTag` and assert `tags` after create.

PR #10425, opened 2026-09-07
resource/alicloud_ecs_snapshot: add configurable wait_until policy

api-tool-agent:
Add the provider-local `wait_until` argument to `alicloud_ecs_snapshot` to select the completion condition for **creation only**, and expose the ECS `available` value as a computed boolean.

- Omitting `wait_until` still waits for `Status=accomplished`. The schema has no default: the runtime `GetOk` fallback preserves existing creation behavior without inserting `wait_until=accomplished` into legacy state or creating a default-value configuration diff after refresh. Explicit `accomplished` and `available` are validated; an explicitly empty string is not a valid policy.
- Explicit `wait_until = "available"` waits for `Available=true`; background upload may still be in progress, and Read preserves the actual `status`. At `Available=true`, the snapshot supports disk creation, disk rollback, snapshot sharing, and image creation, subject to each operation's other requirements.
- `available` is read-only, populated on every Read, and independent of the waiting policy. A reported false replaces a previous true; an absent/null field clears it to false. Import refresh populates `available` but cannot recover `wait_until`.
- `wait_until` only controls when Terraform finishes waiting for snapshot creation; it does not change snapshot capabilities. Neither the local policy nor the computed availability is sent to ECS. Changing or removing only the policy neither recreates nor modifies the snapshot and is not a readiness barrier. Removing an explicitly stored policy remains a normal local configuration change and converges without a residual diff.
- Metadata updates (`snapshot_name`, legacy `name`, and `description`) no longer wait for snapshot upload after a successful `ModifySnapshotAttribute` call; they finish with the normal Read. Live probes showed these updates already persisted while the snapshot was still uploading. This intentionally removes the previous Update completion wait rather than treating upload status as an attribute-update result.
- Retention changes, including mixed metadata/retention requests, still wait for `Status=accomplished` **before** the API call in either mode. This precondition is backed by observed `InvalidSnapshotId.NotReady` responses followed by successful identical requests after upload completed. There is no extra post-update wait.
- Resource-group updates have no extra snapshot-readiness wait; tag handling is unchanged. `wait_until` does not govern Update operations.

Retain the five existing acceptance cases covering names, retention, resource groups, tags, import and destroy, including omitted/explicit accomplished creation and an available-mode snapshot consumed by a dependent disk. They now assert the computed availability, with omitted creation also asserting that no accomplished policy is written into state. Tests use compatible encrypted ESSD fixtures in matching zones. `AliCloudEcsSnapshotBasicDependence1` remains as a wrapper around the shared fixture; `basic1` and `basic1_twin` are retained, but their original `cloud_efficiency` disk-type coverage is no longer exercised. The generic category case uses `standard`; schema and RPC round-trip coverage retain legacy `flash` without claiming live cloud lifecycle coverage for it. The documented capabilities do not imply that this acceptance suite exercises all four downstream operations.

Validation for the latest schema/output revision:

- Tests-first regressions failed against the old implementation, exposing the legacy default-value diff, the swallowed removal of an accomplished policy, and the missing availability output. After the patch, all nine Snapshot-specific groups passed: 31 named subcases plus the schema check and the six-step import/Read sequence, with zero failures or skips (125.657s). Command: `GOMAXPROCS=2 go test -p=1 -vet=off ./alicloud -run '^TestUnitEcsSnapshot' -count=1 -v`.
- Coverage includes both creation policies, omitted-policy waiting and state, normal legacy Refresh→Diff compatibility, policy changes/removals and convergence, import, true→false/missing/null refreshes, metadata and legacy-name updates, retention-only/mixed preconditions, resource-group updates, and legacy category RPC round trips. No new test framework or cloud probe was added.
- Documentation content, Markdown lint, formatting and diff checks passed. The repository coverage checker was compiled against the updated schema and reported 100% attribute testing coverage and 100% modification coverage; this is schema/config coverage, not a live acceptance result.
- The single `go vet -p=1 ./alicloud` run reports only the two existing unused `fmt.Sprintln` results in unchanged `common.go` (lines 1592 and 1604); vet is not reported as clean.

Remote acceptance and all reported PR checks passed on commit `1d729f66a39b51b53deb45612b12858be2860b73`. The actual acceptance log confirms all five cases (`basic0`, `basic0_twin`, `basic1`, `basic1_twin`, and `availableDisk`) passed, with zero failures or skips. This run includes the schema-default removal and computed availability output, not an older implementation.

The latest revision only adds `Available since v1.293.0` to the `wait_until` documentation, matching the current unreleased development version (v1.292.0 is already released). Documentation content, Markdown lint and diff checks passed again. Provider and test code are unchanged from the acceptance-tested commit above; no new acceptance run was manually dispatched for this documentation-only revision. Checks on the new documentation-only head have not yet been verified.

### DIFF
--- a/alicloud/resource_alicloud_ecs_snapshot.go
+++ b/alicloud/resource_alicloud_ecs_snapshot.go
@@ -1,4 +1,3 @@
-// Package alicloud. This file is generated automatically. Please do not modify it manually, thank you!
 package alicloud
 
 import (
@@ -27,6 +26,10 @@ func resourceAliCloudEcsSnapshot() *schema.Resource {
 			Delete: schema.DefaultTimeout(5 * time.Minute),
 		},
 		Schema: map[string]*schema.Schema{
+			"available": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
 			"category": {
 				Type:         schema.TypeString,
 				Optional:     true,
@@ -84,6 +87,11 @@ func resourceAliCloudEcsSnapshot() *schema.Resource {
 				Computed: true,
 			},
 			"tags": tagsSchema(),
+			"wait_until": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: StringInSlice([]string{"accomplished", "available"}, false),
+			},
 			"name": {
 				Type:          schema.TypeString,
 				Optional:      true,
@@ -152,12 +160,20 @@ func resourceAliCloudEcsSnapshotCreate(d *schema.ResourceData, meta interface{})
 	d.SetId(fmt.Sprint(response["SnapshotId"]))
 
 	ecsServiceV2 := EcsServiceV2{client}
-	stateConf := BuildStateConf([]string{}, []string{"accomplished"}, d.Timeout(schema.TimeoutCreate), 5*time.Second, ecsServiceV2.EcsSnapshotStateRefreshFunc(d.Id(), "Status", []string{}))
+	stateField, target := ecsSnapshotWaitCondition(d)
+	stateConf := BuildStateConf([]string{}, []string{target}, d.Timeout(schema.TimeoutCreate), 5*time.Second, ecsServiceV2.EcsSnapshotStateRefreshFunc(d.Id(), stateField, []string{}))
 	if _, err := stateConf.WaitForState(); err != nil {
 		return WrapErrorf(err, IdMsg, d.Id())
 	}
 
 	return resourceAliCloudEcsSnapshotRead(d, meta)
+}
+
+func ecsSnapshotWaitCondition(d *schema.ResourceData) (string, string) {
+	if v, ok := d.GetOk("wait_until"); ok && v == "available" {
+		return "Available", "true"
+	}
+	return "Status", "accomplished"
 }
 
 func resourceAliCloudEcsSnapshotRead(d *schema.ResourceData, meta interface{}) error {
@@ -174,6 +190,13 @@ func resourceAliCloudEcsSnapshotRead(d *schema.ResourceData, meta interface{}) e
 		return WrapError(err)
 	}
 
+	available := objectRaw["Available"]
+	if available == nil {
+		available = false
+	}
+	if err := d.Set("available", available); err != nil {
+		return WrapError(err)
+	}
 	if objectRaw["Category"] != nil {
 		d.Set("category", objectRaw["Category"])
 	}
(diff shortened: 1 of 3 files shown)


gzzzsGit (approved):

gzzzsGit (approved):

gzzzsGit (approved):

xuzhang3 (approved):
The waiting semantics now track the server's own readiness boundary, and each path is backed by evidence: live probes showing metadata/resource-group updates succeed while `Available=true` and `Status=progressing`, while retention changes are rejected with `InvalidSnapshotId.NotReady` until upload completes; per-path RPC regressions (including the no-pre-wait assertion for metadata updates and the wait-then-modify ordering for retention); and an acceptance test that creates a disk from the still-uploading snapshot and verifies `SourceSnapshotId` through the API. Keeping `Read`'s actual status (possibly `progressing` right after apply) and documenting the behavior in the docs NOTE is the right call.

For the record, no action needed: apply now completes while background upload may still be in progress and downstream resources proceed earlier — this matches the API's availability contract, is PATCH-level (no schema/ID/attribute changes), and is documented.

gzzzsGit (approved):

alicloud/resource_alicloud_ecs_snapshot.go
```diff
@@ -287,6 +312,11 @@ func resourceAliCloudEcsSnapshotUpdate(d *schema.ResourceData, meta interface{})
 
 	request["ResourceType"] = "snapshot"
 	if update {
+		stateConf := BuildStateConf([]string{}, []string{"true"}, d.Timeout(schema.TimeoutUpdate), 5*time.Second, ecsServiceV2.EcsSnapshotStateRefreshFunc(d.Id(), "Available", []string{}))
```
gzzzsGit: 这段是否必要？

alicloud/resource_alicloud_ecs_snapshot.go
```diff
@@ -266,8 +291,8 @@ func resourceAliCloudEcsSnapshotUpdate(d *schema.ResourceData, meta interface{})
 		if err != nil {
 			return WrapErrorf(err, DefaultErrorMsg, d.Id(), action, AlibabaCloudSdkGoERROR)
 		}
-		ecsServiceV2 := EcsServiceV2{client}
-		stateConf := BuildStateConf([]string{}, []string{"accomplished"}, d.Timeout(schema.TimeoutUpdate), 5*time.Second, ecsServiceV2.EcsSnapshotStateRefreshFunc(d.Id(), "Status", []string{}))
+		stateField, target := ecsSnapshotWaitCondition(d)
```
gzzzsGit: 这里的等待是否必要

gzzzsGit (approved):

alicloud/resource_alicloud_ecs_snapshot.go
gzzzsGit: 本次改动不是生成器生成的，去掉自动生成标记吧

gzzzsGit (approved):

gzzzsGit (approved):

gzzzsGit (approved):

gzzzsGit (approved):

PR #10427, opened 2026-09-07
resource/alicloud_nlb_listener: filter nil/empty elements in ca_certificate_ids and certificate_ids to avoid SDK panic

api-tool-agent:
## Summary

`alicloud_nlb_listener` Create and Update forwarded the raw `[]interface{}` for `certificate_ids` and `ca_certificate_ids` straight to the RPC request without filtering nil or empty-string elements. When a list contained a nil element, tea-rpc-utils panicked with `reflect: Value.Interface: cannot return zero Value` because it reflects over slice elements and the zero Value has no concrete type to recover.

## Changes

- Add `filterNilAndEmptyStrings` helper in `alicloud/common.go` (next to the existing `filterEmptyStrings`) that drops nil and empty-string entries from a `[]interface{}`.
- Wrap the `CertificateIds` and `CaCertificateIds` assignments in both `resourceAliCloudNlbListenerCreate` and `resourceAliCloudNlbListenerUpdate` with the new helper so the SDK serializer never receives a slice containing nil elements.

## Validation

- `go vet -p=1 ./alicloud` (package-scoped): clean for the changed code. Two pre-existing `fmt.Sprintln` warnings at `common.go:1612/1624` are unrelated master debt.
- Remote acceptance tests pending.

### DIFF
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -414,6 +414,26 @@ func filterEmptyStrings(arr []interface{}) []interface{} {
 	return result
 }
 
+// filterNilAndEmptyStrings removes nil and empty-string elements from a
+// []interface{} before it is forwarded to the SDK request. The tea-rpc-utils
+// request serializer reflects over slice elements; a nil element triggers
+// "reflect: Value.Interface: cannot return zero Value" because the zero
+// Value has no concrete type to recover. Empty strings are dropped as well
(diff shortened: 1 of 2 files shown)


xuzhang3 (changes requested):
在 schemm  补充validation，限制 nil 或者 空字符串的进入

PR #10455, opened 2026-09-08
resource/alicloud_ram_role_policy_attachment: support resource group scope

api-tool-agent:
Add optional `resource_group_id` support to `alicloud_ram_role_policy_attachment`.

- Send the selected scope in RAM attach/detach requests. Read through `RamServiceV2` using ResourceManager `ListPolicyAttachments`, explicit scope, and exact attachment matching.
- Preserve existing four-part account IDs and add five-part resource-group IDs for imports. Read and delete use the stored identity; pagination, visibility delays, and error handling preserve scope and state.
- Document the additional `ram:ListPolicyAttachments` read permission and account/resource-group coexistence. Add a scoped acceptance case and retain the Custom-policy and read-only System-policy cases.

Validation: Go 1.25.0 acceptance-test compilation and `go vet -p=1 ./alicloud` passed for the service-call cleanup. The three acceptance tests passed at `99bab17e4fc9fc5fd7b8b3e17a56561fc06552a9` (39.91s, 6.10s, 6.07s), covering account/A/B coexistence, all three imports, independent deletion, scope replacement and clearing, and omitted/empty scope equivalence. The acceptance-test source is unchanged; the later service-call cleanup preserves the validated query and lifecycle behavior.

Also removes two no-op `fmt.Sprintln` calls that block package vet, preserving the existing YAML comparison behavior.

### DIFF
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -1587,10 +1587,7 @@ func compareCmsHybridMonitorFcTaskYamlConfigAreEquivalent(tem1, tem2 string) (bo
 	}
 
 	var P1 Products
-	err := yaml.Unmarshal([]byte(tem1), &P1)
-	if err != nil {
-		fmt.Sprintln(false)
-	}
+	_ = yaml.Unmarshal([]byte(tem1), &P1)
 
 	y1 := make([]string, 0)
 	for _, product := range P1.Products {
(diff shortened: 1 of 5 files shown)


gzzzsGit (approved):

alicloud/common.go
```diff
@@ -1587,10 +1587,7 @@ func compareCmsHybridMonitorFcTaskYamlConfigAreEquivalent(tem1, tem2 string) (bo
 	}
 
 	var P1 Products
-	err := yaml.Unmarshal([]byte(tem1), &P1)
-	if err != nil {
-		fmt.Sprintln(false)
-	}
+	_ = yaml.Unmarshal([]byte(tem1), &P1)
```
gzzzsGit: 这个修改的意义是什么

gzzzsGit (approved):

gzzzsGit (approved):

gzzzsGit (approved):

PR #10480, opened 2026-09-09
resource/alicloud_drds_instance: read complete instance attributes and retry transient errors

api-tool-agent:
Importing a DRDS instance omitted required attributes and could produce a replacement plan. A newly created instance can also reach `RUN` while its specification or network fields are incomplete.

This change reads the required attributes and charge type, exposes `status`, and waits for complete metadata before updating state on create, import, and refresh. Transient Describe errors are retried without treating the instance as deleted. VPC and VSwitch IDs are read from the same VIP.

It also preserves the API's `Dns` response field despite the SDK's case-sensitive `dns` tag, while retaining the SDK's numeric port decoding. Acceptance fixtures use the supported sn2 series, and the VPC lifecycle case is pinned to Hangzhou.

Validation:

- Raw-JSON regressions reproduce empty connection strings before the DNS fix and pass afterward, including lowercase `dns`, numeric ports, and a public VIP preceding the intranet VIP.
- The VPC lifecycle and all three regression test functions pass remotely in Hangzhou: four PASS, zero FAIL/SKIP. The lifecycle covers creation, required fields and DNS readback, import-state attribute comparison, a same-configuration plan, description update/restore, and Provider destruction with the framework's destroy check.
- A separate real Terraform 1.9.7 CLI test also passes using production source at commit `85f690b6381888bb56a1180b96d1c8bdc77857ee`. An isolated directory with no state imports a new Hangzhou PostPaid/sn2 instance by ID, without copying its creation state. Both `plan -refresh=false -detailed-exitcode` and `plan -refresh=true -detailed-exitcode` return 0, and the saved plan JSON reports `["no-op"]` for the instance. The creation directory then destroys the fixture successfully and has no managed resources remaining.
- The separate CLI harness is a verification-only fixture, not part of this PR. It uses the compiled provider under test rather than a registry provider. This is distinct from the SDK import-state comparison, which does not persist the imported state for subsequent steps.
- Attribute and modification coverage both pass at 100%.
- Local package vet still reports two pre-existing unused `fmt.Sprintln` results in unchanged `common.go`.
- Earlier Beijing lifecycle failures are not the current Hangzhou result.

The final VPC test-file changes after the first remote validation only indent the embedded provider configuration and make the existing `ExpectNonEmptyPlan: false` default explicit for the repository's coverage parser; no resource behavior, test values, or assertions were relaxed. The independent empty-state CLI test validates the pushed production code.

Scope: one Hangzhou PostPaid/sn2 fixture, Terraform 1.9.7 on Linux amd64. DNS field readback is verified; DNS resolution and database connectivity are not part of this test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 ## 1.293.0 (Unreleased)
+
+BUG FIXES:
+
+- resource/alicloud_drds_instance: fix read not returning the required attributes vswitch_id, specification and instance_series, and add a computed status attribute. ([#10480](https://github.com/aliyun/terraform-provider-alicloud/pull/10480))
 ## 1.292.0 (September 8, 2026)
 
 - **New Resource:** `alicloud_threat_detection_rd_default_sync_list` ([#10207](https://github.com/aliyun/terraform-provider-alicloud/issues/10207))
--- a/alicloud/drds_instance_read_test.go
+++ b/alicloud/drds_instance_read_test.go
@@ -1,0 +1,169 @@
+package alicloud
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	sdkendpoints "github.com/aliyun/alibaba-cloud-sdk-go/sdk/endpoints"
+	"github.com/aliyun/credentials-go/credentials"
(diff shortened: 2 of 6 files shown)


gzzzsGit (approved):

api-tool-agent:
Branch was rebased onto current master to resolve the changelog conflict; continued in #10520 (same fork branch, same change).

PR #10476, opened 2026-09-09
feat(dts): support source_endpoint_ssl for subscription job

api-tool-agent:
## Summary

Add `source_endpoint_ssl` to `alicloud_dts_subscription_job` (DTS `SourceEndpoint.SslSolutionEnum`), matching the existing migration/synchronization job resources (v1.287.0 #10099).

## Changes

- Schema: `source_endpoint_ssl` Optional+Computed, ValidateFunc `["0","1"]`
- Create/Update: fold srcSSL into `ConfigureSubscription.Reserve` via `setDtsEndpointSSL` (merged with user reserve, not overwritten); Update adds SetPartial
- Read/Import: backfill via `convertDtsEndpointSslResponse`
- Docs: `website/docs/r/dts_subscription_job.html.markdown`
- CHANGELOG: 1.293.0 (Unreleased) ENHANCEMENTS entry
- Test: `TestAccAliCloudDTSSubscriptionJob_basic0` extended with 1→0→1 in-place update + ImportState verification

## Verification

- gofmt: clean
- go vet: no new warnings
- `go test -run '^$' ./alicloud`: pass (package compiles)
- check-public-diff scan: pass
- ACC `TestAccAliCloudDTSSubscriptionJob_basic0` (C1-C4/C6): pending QA execution

baseCommit: 57b19059ad55a38ccc632bde38ac11430f6dbf56
headCommit: 87fc3e2e8d03b1369806796850e6e25b5823197e

PR #10487, opened 2026-09-09
resource/alicloud_dts_subscription_job: support source_endpoint_ssl; testcase: fix acceptance test region resolution

api-tool-agent:
## Summary

Support `source_endpoint_ssl` for `alicloud_dts_subscription_job` (DTS `SourceEndpoint.SslSolutionEnum`), mirroring the existing migration/synchronization job resources, and fix the pre-existing acceptance test defect that reads `ALICLOUD_REGION` before `testAccPreCheck` applies its cn-beijing fallback.

## Changes

- `resource/alicloud_dts_subscription_job`: add `source_endpoint_ssl` (Optional+Computed, valid values `0`/`1`) to the Schema; read back from `SslSolutionEnum`; write through `ConfigureSubscription.Reserve` on Create/Update (merged with the user `reserve` map, never overwritten); ImportState passthrough with ImportStateVerify.
- `testcase`: in `resource_alicloud_dts_subscription_job_test.go`, replace 6 construct-time `os.Getenv("ALICLOUD_REGION")` reads (Config/Check maps of basic0/basic1/basic2 and `AliCloudDTSSubscriptionJobBasicDependence0`) plus 3 name renders with a `dtsSubscriptionRegion()` helper whose cn-beijing fallback mirrors `testAccPreCheck`. This fixes Step 0 `CreateDtsInstance` returning 403 `MissingSourceRegion` when the test runner has no `ALICLOUD_REGION` set.
- docs + CHANGELOG updated for the new field.

## Tests

- Targeted acceptance test: `TestAccAliCloudDTSSubscriptionJob_basic0` (covers ssl create/read, 1→0→1 in-place update, ImportState).
- Regression scope: basic0/basic1/basic2 share the region fix; the existing main case is basic0.
- Local self-checks run: gofmt clean; `go vet -p=1 ./alicloud` shows only 2 pre-existing baseline warnings in `common.go` (untouched); alicloud package compiles (`go test -run '^$'`); public-content diff scan passes. Remote acceptance tests are executed by the QA pipeline, not locally.

## Example

No example changes; documentation reference updated.

Supersedes #10476 (same change on an older base, without the test fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 ## 1.293.0 (Unreleased)
+
+ENHANCEMENTS:
+
+- resource/alicloud_dts_subscription_job: support source_endpoint_ssl.
 ## 1.292.0 (September 8, 2026)
 
 - **New Resource:** `alicloud_threat_detection_rd_default_sync_list` ([#10207](https://github.com/aliyun/terraform-provider-alicloud/issues/10207))
--- a/alicloud/resource_alicloud_dts_subscription_job.go
+++ b/alicloud/resource_alicloud_dts_subscription_job.go
@@ -155,6 +155,12 @@ func resourceAliCloudDtsSubscriptionJob() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"source_endpoint_ssl": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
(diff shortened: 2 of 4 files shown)


api-tool-agent:
Superseded by #10490, which adds the acceptance-test fix for the precheck failure (source RDS SSL encryption) and fast-fails on PrecheckFailed in the status flow. Closing this PR.

PR #10491, opened 2026-09-09
resource/alicloud_vpc_dhcp_options_set_attachment: fix delete state refresh to terminate when the association is gone

api-tool-agent:
Fixes the Delete state refresh of alicloud_vpc_dhcp_options_set_attachment:

1. Delete now queries GetDhcpOptionsSet with only the dopt-* options set id (not the composite vpc_id:dopt_id value).
2. The delete wait terminates when the association is gone, so deleting an already-disassociated attachment no longer times out.

Changes:
- resource_alicloud_vpc_dhcp_options_set_attachment.go: use the delete-specific state refresh on Delete.
- service_alicloud_vpc.go: add DescribeVpcDhcpOptionsSetAttachmentDeleteStateRefreshFunc, returning nil when the association no longer exists.
- resource_alicloud_vpc_dhcp_options_set_attachment_test.go: rename the acceptance test to the TestAccAliCloud prefix and add unit tests for refresh id handling, association-gone termination, and associated-state polling.

Testing:
- Unit: regression tests pass locally.
- ACC: remote acceptance tests passed against identical content.

Supersedes #10488.


api-tool-agent:
Superseded by #10510: adds TimeoutDelete fix per review.

PR #10495, opened 2026-09-10
resource/alicloud_ess_alarm: add Throttling.User retry to Read, Update, and Delete paths

api-tool-agent:
## Summary

- Add throttling retry with exponential backoff to `alicloud_ess_alarm` resource Read, Update, and Delete paths
- Unify all write-path retry criteria on `NeedRetry(err)` (matches old SDK `Throttling.User`, `ServiceUnavailable`, 5xx and network errors)
- Add missing retry to the `DisableAlarm` call that follows `CreateAlarm` when `enable = false`

### service_alicloud_ess.go — `DescribeEssAlarm`
Wrap both `DescribeAlarms` API calls (system + custom metric types) with `resource.Retry(5*time.Minute)` + `incrementalWait(3s,3s)` + `NeedRetry(err)`. `NeedRetry` handles old SDK `*errors.ServerError` with regex matching on `"Throttling"` (matches `"Throttling.User"`).

### resource_alicloud_ess_alarm.go — `Create`
- `CreateAlarm` retry criteria unified to `NeedRetry(err)` so old SDK `ServerError` with `ErrorCode: "Throttling.User"` is properly matched
- `DisableAlarm` (executed right after `CreateAlarm` when `enable = false`) wrapped with `resource.Retry(5*time.Minute)` + `NeedRetry(err)`

### resource_alicloud_ess_alarm.go — `Update`
Add `resource.Retry(5*time.Minute)` + `NeedRetry(err)` around `ModifyAlarm`, `EnableAlarm`, and `DisableAlarm` API calls.

### resource_alicloud_ess_alarm.go — `Delete`
Add `resource.Retry(5*time.Minute)` + `NeedRetry(err)` around `DeleteAlarm`. 404 is not matched by `NeedRetry`, so it fails fast and is handled by the existing 404 -> remove-from-state path.

## Tests

- Acceptance tests for `alicloud_ess_alarm`: 6/6 PASS, zero regressions. Covered: `TestAccAliCloudEssAlarmBasic`, `TestAccAliCloudEssAlarmWithExpression`, `TestAccAliCloudEssAlarmWithEffective`, `TestAccAliCloudEssAlarmWithEffectiveModify` (enable flip), `TestAccAliCloudEssAlarmMulti`.
- `go build ./alicloud/` PASS, `gofmt` clean (CI Formatter gate), `go vet` shows no new warnings.

## Known tradeoff

[...]

PR #10503, opened 2026-09-10
resource/alicloud_ess_alarm: add Throttling.User retry to Read, Update, and Delete paths

api-tool-agent:
## Summary

- Add throttling retry with exponential backoff to `alicloud_ess_alarm` resource Read, Update, and Delete paths
- Unify all write-path retry criteria on `NeedRetry(err)` (matches old SDK `Throttling.User`, `ServiceUnavailable`, 5xx and network errors)
- Add missing retry to the `DisableAlarm` call that follows `CreateAlarm` when `enable = false`

### service_alicloud_ess.go — `DescribeEssAlarm`
Wrap both `DescribeAlarms` API calls (system + custom metric types) with `resource.Retry(5*time.Minute)` + `incrementalWait(3s,3s)` + `NeedRetry(err)`. `NeedRetry` handles old SDK `*errors.ServerError` with regex matching on `"Throttling"` (matches `"Throttling.User"`).

### resource_alicloud_ess_alarm.go — `Create`
- `CreateAlarm` retry criteria unified to `NeedRetry(err)` so old SDK `ServerError` with `ErrorCode: "Throttling.User"` is properly matched
- `DisableAlarm` (executed right after `CreateAlarm` when `enable = false`) wrapped with `resource.Retry(5*time.Minute)` + `NeedRetry(err)`

### resource_alicloud_ess_alarm.go — `Update`
Add `resource.Retry(5*time.Minute)` + `NeedRetry(err)` around `ModifyAlarm`, `EnableAlarm`, and `DisableAlarm` API calls.

### resource_alicloud_ess_alarm.go — `Delete`
Add `resource.Retry(5*time.Minute)` + `NeedRetry(err)` around `DeleteAlarm`. 404 is not matched by `NeedRetry`, so it fails fast and is handled by the existing 404 -> remove-from-state path.

## Tests

- Acceptance tests for `alicloud_ess_alarm` (QA ACC taskId 10951, 6/6 PASS on the prior revision): `TestAccAliCloudEssAlarmBasic`, `TestAccAliCloudEssAlarmWithExpression`, `TestAccAliCloudEssAlarmWithEffective`, `TestAccAliCloudEssAlarmWithEffectiveModify` (enable flip), `TestAccAliCloudEssAlarmMulti`. Round-2 unified the write-path retry criteria, so formal re-verification on this delivery revision is pending.
- `go build ./alicloud/` PASS, `gofmt` clean (CI Formatter gate), `go vet` shows no new warnings.

## Known tradeoff

Each of the two `DescribeAlarms` calls in the Read path can retry up to 5 minutes, and the delete-state wait checks its deadline only after a describe returns, so a sustained throttle can make `terraform destroy` exceed the configured timeout. This is a deliberate choice to keep the repository-wide 5-minute retry convention; a shorter read budget can be applied later if needed.

## Pattern Reference
- `resource.Retry(5*time.Minute)` + `NeedRetry(err)` is the mainstream convention in this repository
- Other ESS resources: `ess_scaling_group`, `ess_scaling_configuration`; NAS access rule retry (commit 93c4601f6)

## Replaces

This PR replaces #10495. The previous delivery branch carried 2 commits and failed the upstream "Pull Request Max Commits" policy (commit count must be ≤ 1). This PR delivers the identical tree as a single commit with the same author identity, and includes the review fixes applied in #10495. Review conclusions and discussion history: https://github.com/aliyun/terraform-provider-alicloud/pull/10495.

### DIFF
--- a/alicloud/resource_alicloud_ess_alarm.go
+++ b/alicloud/resource_alicloud_ess_alarm.go
@@ -172,7 +172,7 @@ func resourceAliyunEssAlarmCreate(d *schema.ResourceData, meta interface{}) erro
 			return essClient.CreateAlarm(request)
 		})
 		if err != nil {
-			if IsExpectedErrors(err, []string{Throttling}) {
+			if NeedRetry(err) {
 				return resource.RetryableError(err)
 			}
 			return resource.NonRetryableError(err)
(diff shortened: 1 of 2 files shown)


xuzhang3:
Review notes on head `f2c169fe9`:

- No blocking findings.
- The retry additions are consistent with the existing error classification: `NeedRetry` is a superset of the previous Throttling-only match (old-SDK `Throttling.User` is covered; 404/4xx are not), so non-matching errors still fall through to the existing `IsExpectedErrors` checks — including Delete's 404 fast-exit, which keeps failing fast instead of retrying.
- The Read not-found path is synthesized from empty Describe results, outside the retry scope, so state removal on deletion stays intact.
- Both callers of the shared `DescribeEssAlarm` (Read and the wait loop) were re-checked: the only observable effect is the timeout overshoot under sustained throttling, which the PR body documents as a deliberate tradeoff — a latency cost, not a correctness break.
- All CI checks are green.

PR #10505, opened 2026-09-10
resource/alicloud_oos_parameter: support the write-only attribute value_wo

xuzhang3:
### Description / Why

`alicloud_oos_parameter` persists the parameter value in the state and plan, which is undesirable when the parameter holds a secret (database password, token). This PR adds write-only support so the value can be managed without ever being stored:

- `value_wo` - (Optional, Sensitive) write-only twin of `value`: sent to the server on create and update, never stored in the state or plan. Carries the same `1` to `4096` length validation as `value`, so an invalid value fails at plan time instead of at apply. Requires Terraform >= 1.11.
- `value_wo_version` - (Optional, Computed) change trigger for the write-only value: write-only values are invisible to the diff, so the value is re-sent only when this argument changes. After both are removed from the configuration, the last version remains in the state.
- `has_value_wo` - (Computed) reports whether the value is currently managed by `value_wo`.
- `value` is relaxed from Required to Optional + Computed with `ExactlyOneOf` between `value` and `value_wo`; while `value_wo` is in use, the value read back from the API is not persisted and `value` reads empty.

### Changes

- `alicloud/resource_alicloud_oos_parameter.go`: new schema entries, write-only value resolution from the raw configuration in Create/Update, read-side mode handling for `has_value_wo` and `value`.
- `alicloud/common.go`: `getWriteOnlyStringValue` helper that reads a write-only attribute from the raw configuration (write-only values exist nowhere else).
- `alicloud/resource_alicloud_oos_parameter_test.go`: new `TestAccAliCloudOosParameter_valueWo` case; renames `TestAccAlicloudOOSParameter_basic0/basic1` to `TestAccAliCloudOosParameter_basic0/basic1` (fixture repair, see below).
- `website/docs/r/oos_parameter.html.markdown`: argument/attribute documentation plus a note on the Terraform version requirement and read-back behavior.
- `CHANGELOG.md`: entry under 1.293.0.

#### Test visibility repair

The existing cases `TestAccAlicloudOOSParameter_basic0/basic1` were invisible to the acceptance runner's case-sensitive `TestAccAliCloud{Namespace}{Resource}*` pattern (legacy `Alicloud` casing and the `OOS` abbreviation), so they never executed remotely. They are renamed to `TestAccAliCloudOosParameter_basic0/basic1` to match the convention used by the other resources. No test logic changed beyond the names.

### Acceptance tests

Pattern `TestAccAliCloudOosParameter*` — 3 pass / 0 fail / 0 skip:

```
PASS: TestAccAliCloudOosParameter_basic0 (51.28s)
PASS: TestAccAliCloudOosParameter_basic1 (10.65s)
PASS: TestAccAliCloudOosParameter_valueWo (19.65s)
```

`valueWo` covers the full lifecycle: create with `value_wo` + version 1 (state: `value` empty, `has_value_wo` true, `value_wo_version` 1), rotate the value with version 2, remove both from the configuration (asserts the last version is retained in the state and no update fires), then import with `value_wo_version` in `ImportStateVerifyIgnore`.

Coverage gap: the opposite migration direction — an existing `value` configuration switching to `value_wo` — uses the same update path (the `value_wo_version` diff triggers the update and the write-only value takes precedence in the request), but has no dedicated acceptance step; the end-to-end covered direction is write-only to legacy.

### Notes for reviewers

- `value_wo_version` is Optional + Computed rather than plain Optional: with a plain Optional `TypeInt`, the SDK zero-materializes a removed version as `0` in the state, which breaks import verification (state carries `0`, import produces absence) and produces a perpetual `0 -> null` diff that re-triggers updates on every plan. Computed semantics retain the last value with no diff, consistent with other Optional + Computed attributes in this codebase.
- The API requires `Value` on every update call, so the resolved value (from `value_wo` when set, falling back to `value`) rides the update request regardless of which argument triggered the update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 ## 1.293.0 (Unreleased)
+
+ENHANCEMENTS:
+
+- resource/alicloud_oos_parameter: support the write-only attribute `value_wo` with `value_wo_version` as its change trigger, so the parameter value can be managed without being persisted in the state or plan.
+
 ## 1.292.0 (September 8, 2026)
 
 - **New Resource:** `alicloud_threat_detection_rd_default_sync_list` ([#10207](https://github.com/aliyun/terraform-provider-alicloud/issues/10207))
--- a/alicloud/common.go
+++ b/alicloud/common.go
@@ -33,6 +33,7 @@ import (
 	"github.com/denverdino/aliyungo/common"
 	"github.com/denverdino/aliyungo/cs"
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -727,6 +728,20 @@ func buildClientToken(action string) string {
 		token = token[0:64]
 	}
 	return token
+}
+
+func getWriteOnlyStringValue(d *schema.ResourceData, path cty.Path) (string, error) {
+	if d.GetRawConfig().IsNull() {
+		return "", nil
+	}
+	value, diags := d.GetRawConfigAt(path)
+	if diags.HasError() {
+		return "", WrapError(fmt.Errorf("error retrieving write-only argument: %s", diags[0].Summary))
+	}
+	if value.IsNull() || !value.IsKnown() || !value.Type().Equals(cty.String) {
+		return "", nil
+	}
+	return value.AsString(), nil
 }
 
 func getNextpageNumber(number requests.Integer) (requests.Integer, error) {
--- a/alicloud/resource_alicloud_oos_parameter.go
+++ b/alicloud/resource_alicloud_oos_parameter.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/aliyun/terraform-provider-alicloud/alicloud/connectivity"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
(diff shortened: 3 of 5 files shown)


xuzhang3:
### Self-review result (head `23d71033`)

Business-semantics review of the write-only change — checked schema shape, CRUD wiring, state/plan behavior, and test evidence against the API contract:

**Schema**
- `value` Required → Optional + Computed with `ExactlyOneOf(value, value_wo)` is a widening change — existing configurations are unaffected.
- `value_wo_version` is Optional + Computed rather than plain Optional: a plain Optional `TypeInt` is zero-materialized on config removal, so the state would carry `"0"` while import produces absence (import-verify diff), and every subsequent plan would re-fire an update on the `0 -> null` diff — observed as server-side version inflation during development. Computed semantics retain the last version with no diff.
- `value_wo` carries the same `StringLenBetween(1, 4096)` as `value` (both feed the same API `Value` field). Validation runs at plan time — write-only stripping applies to plan/state only, not to config validation — and the error message exposes only the length, never the content.

**CRUD**
- `Value` rides every update request unconditionally; the API rejects updates without it, which the acceptance run confirmed on the wire.
- Create and Update terminate in Read; Read clears the ID only on a recognized NotFound.
- While `value_wo` mode is active, Read does not persist the value the API returns (`has_value_wo` marker + setting `value` to null). OOS `GetParameter` returns the value, which places this resource in the SSM class.

**Tests**

```
PASS: TestAccAliCloudOosParameter_basic0 (51.28s)
PASS: TestAccAliCloudOosParameter_basic1 (10.65s)
PASS: TestAccAliCloudOosParameter_valueWo (19.65s)
```

3 pass / 0 fail / 0 skip. The `value_wo` lifecycle case covers create (version 1), rotation (version 2), config removal (version retained in state, no update fired), and import. The legacy → `value_wo` migration direction shares the same update path but has no dedicated step — stated as a coverage gap in the description.

**Result**: 0 open findings; all checks green.

PR #10515, opened 2026-09-11
New Resource: alicloud_ims_user

xuzhang3:
### Description

Adds `alicloud_ims_user` — the first managed resource implemented with terraform-plugin-framework on the release/v2 line. It manages a RAM user through the IMS API (2019-08-15), coexisting with the legacy SDKv2 `alicloud_ram_user` (legacy RAM API line).

### Why is this needed?

release/v2 introduces framework-native resources under `alicloud/service/<product>/`. IMS user is the pilot managed resource for this layout: it establishes the house pattern (schema + validators, patch-style update, typed not-found handling, import) that subsequent v2 resources will follow.

### Changes

- New `alicloud/service/ims/resource_user.go`: full CRUD over CreateUser/GetUser/UpdateUser/DeleteUser. Patch-style `UpdateUser` sends only changed attributes; renaming `user_principal_name` is an in-place update (`NewUserPrincipalName`), with the computed `user_name` following; clearing an optional attribute sends an empty string (same behavior as the legacy `alicloud_ram_user`). Read/Delete treat `EntityNotExist.User` as gone.
- Register `alicloud_ims_user` in `alicloud/service/ims/register.go`.
- New dependency: `terraform-plugin-framework-validators v0.19.0` (first consumer in this PR).
- Consistency checker: skip declarations registered framework-native through the v2 service packages — their schemas are not SDKv2 `ResourcesMap` entries, which failed the gate with `not found in the provider ResourceMap`. SDKv2 resources are still checked as before.
- Docs: `website/docs/r/ims_user.html.markdown`.

### Acceptance tests

PASS: TestAccAliCloudImsUser (17.96s)

Note: the test lives under `alicloud/service/ims/` (v2 layout); the shared AccTest runner currently discovers only top-level `alicloud/` package tests, so this run was executed locally with `TF_ACC=1` against a dedicated test account (region `cn-beijing`). All five steps passed (create, update, rename, clear-to-null, import) plus destroy.

PR #10516, opened 2026-09-11
resource/alicloud_ims_user: wrap API errors with errs.WrapErrorf

xuzhang3:
### Description

`alicloud_ims_user` currently surfaces API-call errors as the raw `RpcPost` error, which renders only the SDK error block (status, code, message, data) — no indication of which provider operation failed or where. This wraps all six API-error diagnostic sites with `errs.WrapErrorf`, so the failing operation and the provider file:line travel with the error.

### Why is this needed?

Error context. A bare SDKError tells the user the API failed but not which CRUD path triggered it. The wrapped form prefixes the operation (`Creating alicloud_ims_user: calling CreateUser`) and the provider source location, matching the house error style used by the SDKv2 resources (every API-call error return goes through `WrapErrorf`).

### Changes

- `alicloud/service/ims/resource_user.go`: wrap the six API-error diag sites (CreateUser, UpdateUser, DeleteUser, and the three GetUser paths) with `errs.WrapErrorf`; import `alicloud/errs`.
- The two assertion diagnostics (missing `User.UserId` in the create response, user not visible via GetUser right after creation) already carry their own context and are unchanged.
- Not-found handling is untouched: `isUserNotExist` inspects the raw error before any wrapping, so `EntityNotExist.User` keeps triggering state removal in Read and the swallow in Delete.

### Acceptance tests

The change only executes on API-error paths, which the happy-path test does not enter; request wiring and every exercised path are unchanged from #10515. Standing evidence from the merged PR: `PASS: TestAccAliCloudImsUser (17.96s)` — local run (TF_ACC=1, cn-beijing); the remote runner does not discover `alicloud/service/*` tests.

### DIFF
--- a/alicloud/service/ims/resource_user.go
+++ b/alicloud/service/ims/resource_user.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/alibabacloud-go/tea/tea"
 	"github.com/aliyun/terraform-provider-alicloud/alicloud/connectivity"
+	"github.com/aliyun/terraform-provider-alicloud/alicloud/errs"
 	"github.com/aliyun/terraform-provider-alicloud/alicloud/provider/fwadapt"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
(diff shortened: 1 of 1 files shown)


xuzhang3:
Review notes on head `985d5019c`:

- No blocking findings.
- All six API-error diagnostic sites now wrap with `errs.WrapErrorf` — the failing operation and provider file:line travel with the error, matching the house error style of the SDKv2 resources.
- Not-found handling is untouched: `isUserNotExist` inspects the raw error before any wrapping (Delete's guard and the GetUser helper's internal check), so `EntityNotExist.User` still drives state removal in Read and the swallow in Delete. The substring fallback would survive a wrapped error anyway.
- No behavior change on success paths: every touched site is an error-only branch; request wiring is unchanged.
- No ACC re-run demanded: the wrap only executes on API-error paths, which the happy-path test does not enter; the standing evidence from #10515 covers the unchanged paths.

PR #10523, opened 2026-09-11
resource/alicloud_vpn_gateway_vpn_attachment: use stable tunnel lists with state migration

api-tool-agent:
Tunnel-only edits can be lost because the set hash only identifies the customer gateway and tunnel index. Including configurable fields in that hash introduces another problem: service-populated optional fields can change the identity of the entire tunnel.

This change uses an ordered tunnel list, matches Read responses by `tunnel_index`, and migrates version-zero state without dropping tunnel IDs or PSKs. PSK, explicit LocalId, and BGP changes are detected independently. Enabling tunnel BGP submits both tunnel BGP configurations in the same update request without resending unchanged IKE or IPsec settings.

PSK updates are serialized only when the PSK itself changes, including for the legacy top-level IKE configuration. Read accepts readable server values so external rotations remain visible; empty or all-asterisk responses retain a previously known value instead of introducing a redacted value after import. Explicitly configured PSKs containing asterisks remain supported. An all-asterisk local secret is retained when the response has a different mask length, is empty, or omits the PSK.

When changing a customer gateway and IKE settings together, an omitted RemoteId is replanned and resolved from the new customer gateway's IP address through DescribeCustomerGateway. Explicit RemoteId values remain intact, including unchanged values. A failed lookup stops the update before ModifyVpnAttachmentAttribute. A gateway-only change does not resubmit unchanged IKE settings.

The service-assigned `role` is response-only in the VPN attachment APIs. Legacy `role` configuration remains accepted with a deprecation warning, but no longer produces unsupported update requests or perpetual diffs. State continues to reflect the service response.

Existing set state is migrated in ascending `tunnel_index` order. Configurations declaring tunnel 2 before tunnel 1 must reorder the complete blocks once before applying. Ordinary updates reject position or identity changes before API calls; whole-resource replacement may establish a new order. The upgrade behavior is documented.

Implementation layout:
- Production migration, ordering, and PSK helpers are consolidated into the resource source file.
- The separate unit-test files are excluded from this PR and retained locally for later packaging and migration validation.

Validation:
- Local SDK and HTTP regression tests passed on the prior revision `b655a9a0ba`, covering legacy state migration, ordering, replacement, masked import, external PSK rotation, isolated BGP/LocalId requests, and explicit PSK changes. These test sources are now retained locally rather than included in the PR; migration testing will be performed during a later local package validation.
- Resource attribute and update coverage checks passed (100%) on `b655a9a0ba`; the acceptance-test source in this revision is unchanged.
- On `b655a9a0ba`, `go vet -unusedresult=false -p=1 ./alicloud` passed; the full-vet baseline has two pre-existing unused `fmt.Sprintln` results in `alicloud/common.go`.
- Local SDK/HTTP tests passed for the RemoteId and all-asterisk PSK fixes, including unknown gateway IDs, explicit unchanged RemoteId, unchanged second-tunnel state, failed lookups, and schema validation. Resource attribute/update coverage remains 100%; `go vet -unusedresult=false -p=1 ./alicloud` passed. Test sources remain outside the PR; migration tests have not been rerun.
- AccTest has deliberately not been run for this revision; it is awaiting another review. The earlier 9 PASS result applies only to commit `9972d71205` and does not validate these PSK changes.

Open merge prerequisites (not resolved by this revision):
- Confirm whether updating IKE settings without a PSK preserves the existing service-side key. The earlier remote acceptance test sent the PSK and does not validate omission semantics.
- Confirm the actual API masking contract, including whether partially masked responses can occur. The current read guard recognizes empty and all-asterisk responses, not partial masking.


alicloud/resource_alicloud_vpn_gateway_vpn_attachment.go
```diff
@@ -13,17 +15,26 @@ import (
...
 
-func vpnTunnelOptionsSpecificationHash(v interface{}) int {
-	m := v.(map[string]interface{})
-	s := fmt.Sprintf("%d-%s", m["tunnel_index"].(int), m["customer_gateway_id"].(string))
-	h := int(crc32.ChecksumIEEE([]byte(s)))
-	if h < 0 {
-		return -h
-	}
-	return h
+func resourceAliCloudVpnGatewayVpnAttachment() *schema.Resource {
```
xuzhang3, line 18: 迁移之后的 schema 保持独立，别在 v0 之上叠加修改。耦合在一起容易埋坑。

xuzhang3 (changes requested):

PR #10546, opened 2026-09-13
resource/alicloud_ram_role_policy_attachment: add resource_group_id field

api-tool-agent:
Add `resource_group_id` field to the RAM Role Policy Attachment resource.

### Changes

- **Schema**: Added `resource_group_id` (Optional, Computed, ForceNew). The `AttachPolicyToRole` API accepts `ResourceGroupId` as an optional query parameter. Since the resource has no Update operation, the field is immutable after creation (ForceNew).
- **Create**: Sends `ResourceGroupId` when the field is set; if omitted, the API uses the default resource group.
- **Read**: Reads back `ResourceGroupId` from the `ListPoliciesForRole` response when available.
- **Docs**: Updated argument reference and attributes reference.
- **Tests**: Added `TestAccAliCloudRamRolePolicyAttachment_basic9052` covering the new field with import verification.

### Test Results

Remote acceptance tests pending.


xuzhang3 (changes requested):
**Verdict: request changes** — `resource_group_id` cannot ship as implemented: with a real value Create fails; with no value the field is never sent (no-op). A write-side parameter alone is not a working feature.

**B1 — A real value breaks Create (read side is RG-blind).** `AttachPolicyToRole` with `ResourceGroupId` creates a *resource-group scoped* authorization, which is a distinct object from the account-level authorization. `ListPoliciesForRole` has no `ResourceGroupId` request parameter and only lists account-level authorizations, so the RG-scoped one is structurally invisible to it (not an eventual-consistency delay). With `resource_group_id` set to any real resource group ID, attach succeeds but the immediate Read gets an empty policy list → `ResourceNotfound` at Step 0 of apply. Two further consequences: `DetachPolicyFromRole` without `ResourceGroupId` cannot remove an RG-scoped authorization (Delete would leave residue in the account), and the doc line "If not specified, the default resource group is used" is wrong — not specifying attaches at the **account level**, a different object, not the default RG.

**B2 — The only test exercising the field is vacuous.** basic9052 gates `resource_group_id` behind `os.Getenv("ALICLOUD_RESOURCE_GROUP_ID")`; when unset the config renders an empty string, the `GetOk` guard skips the send, and the test passes without exercising anything. With the body marked "tests pending", no run so far has ever covered the with-value world — the world in which the resource is broken.

**Fix path — all four pieces are needed.** A verified end-to-end variant:

1. **Read**: resolve RG-scoped authorizations via `ResourceManager` `ListPolicyAttachments` (2020-03-31) — the request filters by `ResourceGroupId` / `PrincipalType=ServiceRole` / `PolicyName`, and the response echoes `ResourceGroupId` per attachment (usable for state backfill).
2. **ID**: extend to `role:<policy_name>:<policy_type>:<role_name>:<resource_group_id>` (5 segments) when the field is set, so the account-level and RG-scoped authorizations of the same (policy, role) pair can coexist as separate resources and import restores the scope. Keep the 4-segment legacy ID unchanged when unset.
3. **Delete**: pass `ResourceGroupId` (5th segment) to `DetachPolicyFromRole`.
4. **Dispatch**: route 5-segment IDs inside `DescribeRamRolePolicyAttachment` to the new resolver — the shared test/check infrastructure resolves existence by that method name, so the routing has to live in the Describe itself.

Gotchas found while verifying: `PrincipalName` in `ListPolicyAttachments` is the full display name (`<role_name>@role.<account_alias>.onaliyunservice.com`), not the bare role name — filter by `PolicyName` and prefix-match `PrincipalName`; account-level records appear in the same view with `ResourceGroupId` equal to the account ID, so distinguish by exact RG ID.

**Evidence after applying all four**: basic9052 reworked to resolve a real resource group via `data.alicloud_resource_manager_resource_groups` (no env gate), import verified with an empty `ImportStateVerifyIgnore`, plus a new coexistence case (account-level + RG-scoped attachments of the same policy/role pair in one config):

PASS: basic9050 (4.55s), basic9051 (4.88s), basic9052 (5.79s), basic9053 (5.84s)

The account-level path (4-segment ID, existing Read/Delete) is unchanged and shows no regression.

PR #10582, opened 2026-09-16
New Resource: alicloud_polardb_gateway_model_service; New Resource: alicloud_polardb_gateway_model_api; New Data Source: alicloud_polardb_gateway_model_services; New Data Source: alicloud_polardb_gateway_model_apis

kittenshow98:
### Description

Add Terraform support for the PolarDB AI gateway model-routing layer:

- add `alicloud_polardb_gateway_model_service` and plural data source
- add `alicloud_polardb_gateway_model_api` and plural data source
- add composite import IDs and raw child ID outputs
- normalize and validate model API route-rules JSON
- redact upstream API keys from provider debug logs and omit them from data sources
- add v1.294.0 documentation and changelog entries

### Tests

- Go package compilation
- targeted unit tests
- tfproviderlint
- resource/document consistency checks
- 100% resource attribute configuration coverage
- 100% resource attribute modification coverage
- breaking-change check

The acceptance configurations are currently PlanOnly. Live OpenAPI contract validation is still required before release.


xuzhang3 (changes requested):
Thank you for the parameter mapping — I verified all eight actions against the public OpenAPI definitions (Create/Modify/Describe/Delete for both ModelService and ModelApi) and found the request mapping accurate, including the `Category`/`ModelCategory` dual-shape read-back, name-based DeleteModelService, and the ApiKey redaction strategy. Two blockers before this can merge:

**Blocker 1: No acceptance-test evidence on the current head.**

The PR body states the acceptance configurations are currently PlanOnly, and there is no `PASS:` record on head `351bfb379`. Beyond the missing evidence, the current test shape cannot produce valid evidence on the remote AccTest runner: `testAccPreCheckPolarDBGatewayAI` gates every test behind `ALICLOUD_POLARDB_ENDPOINT` and `ALICLOUD_POLARDB_GATEWAY_ID` env vars, and the provider block overrides the polardb endpoint from that env — the tests would silently skip. Please attach `PASS: <test name> (<duration>)` runs for all four new resources/data sources on the current head, with the gateway dependency created via the existing `alicloud_polardb_gateway` resource rather than env-injected IDs.

**Blocker 2: Five optional attributes cannot be unset — perpetual diff / silent no-op update.**

`record_input`/`record_output` (`alicloud_polardb_gateway_model_api`) and `input_cost_points_per_million`/`output_cost_points_per_million`/`request_cost_points` (`alicloud_polardb_gateway_model_service`) are only sent when `GetOk` succeeds. Once set, removing the attribute from the configuration produces a plan diff; the update then omits the parameter, the server keeps the old value, and Read restores it — the diff never converges. Per the OpenAPI contract these parameters are optional in `ModifyModelApi`/`ModifyModelService`, so the no-send semantics hold. The "100% modification coverage" only changes values and never removes an attribute, so this path is untested. Either always send the parameters (with live ACC verification of the clearing semantics for an empty value) or document that these attributes cannot be cleared.

Non-blocking notes:

- `polarDBGatewayAIInt` in `polardb_gateway_ai_common.go` appears unused in this change set.
- The PR body mentions changelog entries, but `CHANGELOG.md` is not among the changed files.
- After `terraform import`, `alicloud_polardb_gateway_model_service.api_key` will be empty (the API returns a masked value and Read deliberately keeps the configured value); consider documenting this in the import section.

PR #10602, opened 2026-09-17
resource/alicloud_ssl_certificates_service_certificate: support the write-only attribute key_wo

xuzhang3:
## Summary

Add the write-only attributes `key_wo` and `key_wo_version` to `alicloud_ssl_certificates_service_certificate`, following the pattern already merged for `alicloud_oos_parameter` and `alicloud_kms_ciphertext`:

- `key_wo` (Optional, WriteOnly) — the private key is sent to the API at creation only and never stored in the state or plan; conflicts with `key`; requires `key_wo_version`.
- `key_wo_version` (Optional, ForceNew) — the change trigger: bumping it recreates the resource so the new `key_wo` value is sent.
- While `key_wo` is in use, Read suppresses `key` (reads back empty), so the private key never persists.
- Requires Terraform 1.11.0 or later.

## Breaking change

None. `has_key_wo` was removed, but it was never released.

## Test results

PASS: TestAccAliCloudSslCertificatesServiceCertificate_keyWo (24.01s)
PASS: TestAccAliCloudSslCertificatesServiceCertificate_basic0 (37.38s)
PASS: TestAccAliCloudSslCertificatesServiceCertificate_basic0_twin (11.25s)
PASS: TestAccAliCloudSslCertificatesServiceCertificate_basic1 (37.41s)
PASS: TestAccAliCloudSslCertificatesServiceCertificate_basic1_twin (12.37s)
PASS: TestAccAliCloudSslCertificatesServiceCertificate_basic10998 (36.49s)
PASS: TestAccAliCloudSslCertificatesServiceCertificate_basic10998_twin (11.21s)
PASS: TestAccAliCloudSslCertificatesServiceCertificateApply_basic (44.17s)
PASS: TestAccAliCloudSslCertificatesServiceCertificateValidation_basic (39.67s)
PASS: TestAccAliCloudSslCertificatesServiceCertificateValidation_unissuable (218.04s)
PASS: TestAccAliCloudSslCertificatesServiceCertificateApply_configurationMismatch (14.71s)

`basic1` intermittently fails with "plan was not empty": its config uses `data.alicloud_resource_manager_resource_groups.default.ids.1`, and the listing order is unstable between apply and the post-apply plan check (only `resource_group_id` drifts). It is unrelated to this change and passes on re-run (37.41s above).


xuzhang3:
Reviewed: no blockers.

The write-only pair follows the already-merged pattern: `key_wo`/`key_wo_version` are guarded schema-natively (`ConflictsWith` on `key`, `RequiredWith` in both directions), Create sends `Key` from `key_wo` via the raw-config helper, and Read suppresses `key` while `key_wo_version` is configured, so the private key never persists in state or plan in write-only mode. Create and Update both terminate in Read; the new test covers the full transition path (write-only create → version-bump recreate → removal back to `key` plus import), and all CI gates are green.

One non-blocking note: the suppression is config-presence-based, so importing a certificate that was created in write-only mode reads the key back into state — the same behavior as the azurerm reference design.

PR #10623, opened 2026-09-18
New Resource: alicloud_polardb_polar_fs; New Resource: alicloud_polardb_polar_fs_endpoint; New Data Source: alicloud_polardb_polar_fs_instances; New Data Source: alicloud_polardb_lakebase_tenant_token

kittenshow98:
## Summary

- add `alicloud_polardb_polar_fs` for CreatePolarFs, DescribePolarFsAttribute, and DeletePolarFs
- add `alicloud_polardb_polar_fs_instances` for DescribePolarFs
- add `alicloud_polardb_lakebase_tenant_token` for GetLakebaseTenantToken
- add `alicloud_polardb_polar_fs_endpoint` for PolarFS CreateDBClusterEndpoint, DescribeDBClusterEndpoints, and DeleteDBClusterEndpoint
- add provider registration, v1.294.0 documentation, import support, waiters, and sensitive-value log redaction

## Tests

- targeted provider and PolarFS unit tests
- tfproviderlint
- breaking change check
- documentation consistency check
- testing coverage check: 100% testing and modified coverage for both new resources

Live acceptance tests were not run because they create billed PolarFS resources.


xuzhang3 (changes requested):
I verified all eight actions against the public OpenAPI definitions (CreatePolarFs, DescribePolarFs, DescribePolarFsAttribute, DeletePolarFs, CreateDBClusterEndpoint, DescribeDBClusterEndpoints, DeleteDBClusterEndpoint, GetLakebaseTenantToken) and found the mapping accurate outside the two blockers below — including the top-level `Items` array shape, the endpoint create/delete waiters, the `NeedRetry` retry scoping, the `PolarFsType: "all"` default, and the recursive redaction of `CustomOssAk`/`CustomOssSk`/`Token`/`UserDefaultAccSk`. Two blockers before this can merge:

**Blocker 1: No acceptance-test evidence on the current head.**

Both resources' acceptance tests are `PlanOnly` steps with placeholder IDs (`pc-example`, `pfs-example`, `vpc-example`), and the two data sources ship unit tests only — nothing in this change set has exercised the real APIs. The PR body notes live tests were skipped because PolarFS creates billed resources; a Postpaid instance at minimum spec with one create/destroy cycle would cover both resources and both data sources. Please attach `PASS: <test name> (<duration>)` runs for all four on the current head.

**Blocker 2: `db_type` accepts display names and rejects the actual API values.**

Per the public CreatePolarFs definition, `DBType` takes `MySQL` / `PostgreSQL` (example `MySQL`); `polardb_pg` and `polardb_mysql` appear only as `enumValueTitles` display names, and `polar_agentic_db` does not occur in any of the eight action definitions. This PR's `ValidateFunc` accepts exactly those three display names, and `TestUnitPolarDBPolarFsSensitiveFields` asserts that `MySQL`/`PostgreSQL` must be rejected — the unit test inverts the contract. Even if the server tolerated the create, `DescribePolarFsAttribute` returns `DBType` in the `MySQL`/`PostgreSQL` vocabulary, so Read would write a value the schema forbids back into state, and the ForceNew attribute would force replacement on every apply — a recreate loop.

Non-blocking notes:

- `endpoint_type = "S3Pvtz"` has no contract corroboration: the CreateDBClusterEndpoint doc enum is `Custom`/`DynamoDB` (the `PolarFsInstanceId` parameter exists, so the API reuse itself is fine), and only `Nas`/`S3Gateway` are corroborated via `DescribePolarFsAttribute.EndpointItems`. `address_items.private_zone_connection_string` (`PrivateZoneConnectionString`) is likewise absent from the documented response shape. Please confirm these with the PolarDB side or cover them with real ACC runs.
- The docs list `polardb_pg`/`polardb_mysql`/`polar_agentic_db` as `db_type` valid values — fix together with Blocker 2.

kittenshow98:
image

